Issue #8, opened 2019-08-15
Make resources know what attributes were set


ryanwjackson:
**Is your feature request related to a problem? Please describe.**
When you set an attribute to `nil` (or maybe event `''`), it ignores that value and uses the ledger's current value.  Because of this, you can never _unset_ an attribute in the ledger.

**Describe the solution you'd like**
The resources should know which attributes have been updated (similar to ActiveModel Dirty functions), so that if a value is set to `nil` the resulting API call will include a `null`.

ryanwjackson:
When doing this, we need to ensure that in `temp_app` all of the following succeed assuming that vendor is a required attribute of expense

- Sync a new expense (a.k.a. create) where you reference an existing vendor only by `external_id` and no other attributes.  The vendor should be associated to the expense but not updated.
- Sync an existing expense (a.k.a. update) where you reference an existing vendor only by `external_id` and no other attributes.  The vendor should not be updated.
- Sync an existing expense (a.k.a. update) where the vendor is not supplied at all.

mattgmarcus:
Looks good, thanks Ryan

ryanwjackson:
Ref:
- https://api.rubyonrails.org/classes/ActiveModel/Dirty.html
- https://github.com/rails/rails/blob/66cabeda2c46c582d19738e1318be8d59584cc5b/activemodel/lib/active_model/dirty.rb
- https://github.com/rails/rails/blob/66cabeda2c46c582d19738e1318be8d59584cc5b/activemodel/lib/active_model/attribute_methods.rb#L65
- https://github.com/rails/rails/blob/66cabeda2c46c582d19738e1318be8d59584cc5b/activesupport/lib/active_support/concern.rb

PR #38, opened 2019-09-27
Add resource attribute types and casting based on inputs

### LINKED ISSUE #36: Refactor resource attributes
Resource attributes are defined at the root of the resource, and we run the risk of conflicting with lib-specific attributes like `ledger_id`, `external_id`, `created_at`, etc.

- Enable `resource.attributes` to respond with the resources attributes, excluding `ledger_id` and `external_id`.  So instead of `customer.phone_number`, it would be `customer.attributes.phone_number`.

- Not scoping
- Only handle in serialization.

**Related to**
- #8

### LINKED ISSUE #37: Add attribute type validation and conversion
Introducing types beyond primitives (e.g. `Date`) causes the following issues:
- `Sync` expects a hash of primitives
- Operation validation expects certain types (primitives and other)
- Resources expect correct types (primitives and other)

- Add `type` to resource attributes
- Enable `ResourcesBuilder` to convert to the correct type from a primitive

- Allowing the resource to handle conversion, but IMO this is an annoying side of Rails models, where you cannot disable this functionality on assignment.
- Expecting correct types passed to Sync, but then `rails_app`, or whatever is using the lib, will have to do the conversion and have knowledge of each attribute type.

ryanwjackson:
Fix #37 
Fix #36 (though it doesn't change the input or output, it at least adds a class we can build on top of)

This was mostly needed so we can pass dates properly given they are our first "non-primitive" object type on the resources.

I debated using `ActiveModel`, which would give us the `Dirty` stuff (ref: #8).  My problems are the following:

1. It's a big lib.  We don't know what all we are getting.
2. It's tightly coupled to the DB, though I'm guessing we could probably use it without it.  I saw some very old tutorials, but it didn't seem "easy."
3. I figured it was maybe even more work to try to figure out the inner workings of ActiveModel over just using the parts we need.

@mattgmarcus @jozefvaclavik Thoughts?


lib/ledger_sync/resource_attribute_set.rb
```diff
@@ -0,0 +1,30 @@
...
+
+      false
+    end
+
+    def valid_classes
+      types[type]
+    end
+
+    def self.types
+      @types ||= TYPES
```
mattgmarcus: should this be `@@types` since it's on the class?

lib/ledger_sync/util/resources_builder.rb
```diff
@@ -54,6 +70,38 @@ def build_resource(external_id:, type:)
         )
       end
 
+      def cast_to_date(value)
+        case value
+        when Date
+          pdb
```
mattgmarcus: what's this?

spec/adaptors/operation_contract_spec.rb
```diff
@@ -40,7 +40,7 @@ def operation_paths
     def resource_attributes_for(resource_klass:)
       return [] if resource_klass < LedgerSync::Error
 
-      resource_klass.attributes | [:ledger_id]
+      resource_klass.attributes.keys | [:ledger_id]
```
mattgmarcus: Should this be `|` or `||`?

lib/ledger_sync/resource_attribute_set.rb
ryanwjackson: Yup.  And this probably explains a lot of why I wrote things the way I did...lol

lib/ledger_sync/util/resources_builder.rb
ryanwjackson: Needs removed but see `debug.rb`.

spec/adaptors/operation_contract_spec.rb
ryanwjackson: It's an intersection, so adding the `:ledger_id` key if it doesn't exist in the keys already.

PR #39, opened 2019-09-30
Add dirty tracking for resources

### LINKED ISSUE #8: Make resources know what attributes were set

ryanwjackson:
Fix #8 

This is dependent and based off of #7, so it needs to be merged first and then this base needs shifted to `master`.

Uses ActiveModel::Dirty to track if attributes have changed.  Note: You must call `resource.save` to reset the dirty tracking.  This isn't plugged in anywhere yet.  Just the core code and tests.

@mattgmarcus @jozefvaclavik Let me know if you think anything else needs to go in this.

### DIFF
--- a/spec/resources/dirty_spec.rb
+++ b/spec/resources/dirty_spec.rb
@@ -3,48 +3,64 @@ require 'spec_helper'
 RSpec.describe LedgerSync::Resource do
   let(:resource) { LedgerSync::Customer.new }
 
+
   it do
     expect(resource).not_to be_changed
+    expect(resource.external_id).to be_nil
     resource.external_id = :asdf
+    expect(resource.external_id).to eq(:asdf)
     expect(resource).to be_changed
     expect(resource.changes).to have_key('external_id')
     resource.save
     expect(resource).not_to be_changed
+    expect(resource.external_id).to eq(:asdf)
   end
 
   it do
     expect(resource).not_to be_changed
+    expect(resource.sync_token).to be_nil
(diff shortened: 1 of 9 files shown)


ryanwjackson:
@jozefvaclavik @mattgmarcus I have rebased on `master`.  Can you approve and merge?

ryanwjackson:
Actually wait.  Found a bug.

ryanwjackson:
OK, now good to go.  But please review.  Did some tricky things there.

spec/resources/dirty_spec.rb
```diff
@@ -0,0 +1,50 @@
+require 'spec_helper'
+
+RSpec.describe LedgerSync::Resource do
+  let(:resource) { LedgerSync::Customer.new }
+
+  it do
+    expect(resource).not_to be_changed
+    resource.external_id = :asdf
+    expect(resource).to be_changed
+    expect(resource.changes).to have_key('external_id')
```
mattgmarcus: Maybe test that the change shows up exactly. Like test the key/value pair

ryanwjackson: Added!  Ready to merge when approved.

PR #43, opened 2019-10-03
Add Transfer Resource and Operations

### LINKED ISSUE #23: Add resource support for Transfer (2 hrs)
https://developer.intuit.com/app/developer/qbo/docs/api/accounting/all-entities/transfer

We need to be able to create, update, read, and delete transfers, including using Searchers for this

- Id
- ToAccountRef
- Amount
- FromAccountRef
- PrivateNote
- TxnDate

jozefvaclavik:
Fix #23


mattgmarcus:
Can you just test this out locally and make sure it works if you haven't already? Thanks

PR #44, opened 2019-10-04
Cleanup Upsert

jozefvaclavik:
I would call this as a proof of concept for now.

Currently our `Upsert` operation needs to specify all related operations and then it sets `root_operation` as either `Create` or `Update`. But these two operations knows better what related operations they require (it could be different).

This way we let `Create`/`Update` operation to build necessary required operations and use those instead.

### DIFF
--- a/lib/ledger_sync/adaptors/quickbooks_online/expense/operations/upsert.rb
+++ b/lib/ledger_sync/adaptors/quickbooks_online/expense/operations/upsert.rb
@@ -19,50 +19,26 @@ module LedgerSync
               end
             end
 
-            private
-
             def build
-              op = if qbo_expense?
-                     Update.new(adaptor: adaptor, resource: resource)
-                   else
-                     Create.new(adaptor: adaptor, resource: resource)
-                   end
-
-              build_account_operation(resource.account)
-              resource.line_items.each do |line_item|
-                build_account_operation(line_item.account)
+              op = if find_result.success?
+                Update.new(adaptor: adaptor, resource: resource)
+              else
+                Create.new(adaptor: adaptor, resource: resource)
               end
-              build_vendor_operation
-              add_root_operation(op)
+              op.build
+
+              op.before_operations.each {|o| add_before_operation(o)}
+              op.after_operations.each {|o| add_after_operation(o)}
+              add_root_operation(op.root_operation)
             end
 
-            def build_account_operation(account)
-              account_op = Account::Operations::Upsert.new(
-                adaptor: adaptor,
-                resource: account
-              )
-
-              add_before_operation(account_op)
-            end
-
-            def build_vendor_operation
-              vendor = Vendor::Operations::Upsert.new(
-                adaptor: adaptor,
-                resource: resource.vendor
-              )
-
-              add_before_operation(vendor)
-            end
+            private
 
             def find_result
               @find_result ||= Find.new(
                 adaptor: adaptor,
                 resource: resource
               ).perform
-            end
-
-            def qbo_expense?
-              find_result.success?
             end
           end
         end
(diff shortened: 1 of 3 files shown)


lib/ledger_sync/adaptors/quickbooks_online/expense/operations/upsert.rb
```diff
@@ -19,51 +19,27 @@ class Contract < LedgerSync::Adaptors::Contract
...
-              account_op = Account::Operations::Upsert.new(
-                adaptor: adaptor,
-                resource: account
-              )
+              op.build
 
-              add_before_operation(account_op)
+              op.before_operations.each {|o| add_before_operation(o)}
+              op.after_operations.each {|o| add_after_operation(o)}
+              add_root_operation(op.root_operation)
```
jozefvaclavik: The operation class is `Upsert`, and even previously we would just add before/after operations defined in `Upsert` class and then set `root_operation` either `Create`/`Update`

ryanwjackson:
I generally get what you are saying, but I'm not sure I agree:

1. The [`build` method](https://github.com/LedgerSync/ledger_sync/pull/44/files#diff-d3507f063a50233b424acf526d8f906bR22) was kept `private` in an effort to minimize the public api for operations.
2. The same is currently achieved by the coordinator.  It builds each operation, decomposing them into the necessary primary operations.

I guess I don't see why https://github.com/LedgerSync/ledger_sync/pull/44/files#r331583578 is more desirable than what we have right now.  I would think that we are then just re-writing that code in many operations, so the concept of DRYing it up would lead us towards where we have it now, no?

jozefvaclavik:
I have 3 reasons:
1. why should `Upsert` specify related operations, when `Create` and `Update` knows better what related operations they need.
2. were duplicating code in operation, which sole purpose is to decide if we're going to `Create` or `Update`
3. if we come across scenario that `Create` requires more operations then `Update` we will end up with something like this:
```
def build
  op =  if qbo_payment?
          Update.new(adaptor: adaptor, resource: resource)
        else
          Create.new(adaptor: adaptor, resource: resource)
          build_department_operation <--- operation required only by Create
        end

  build_customer_operation
  add_root_operation(op)
end
```

Also, we're talking about `Null` operation when we have `ledger_id` but we don't provide any `data: {}`. So lets say we pass to ledger
```
payment: {
  pay_1: {
    ledger_id: nil,
    data: {
      amount: 123,
      currency: 'USD',
      invoice: :inv_1,
      line_items: [...]
    }
  }
},
invoice: {
  inv_1: {
    ledger_id: '123',
    data: {}
  }
}
```
We're trying to `Upsert` a `Payment` that references `Invoice` (which by itself is a complicated resource). We start with `Payment::Operations::Upsert` and decide we need to sync Invoice. We create `Invoice::Operations::Upsert` and that defines related operations for `Account`, `Product`, `Customer`, etc.. As we didn't provide any data, we can't sync these resources.

Sure we can trigger their sync only if `resource.customer.present?` or something similar.

Or `Invoice::Operations::Upsert` decides to do `Invoice::Operations::Null` and that operation doesnt trigger any related operations.

lib/ledger_sync/adaptors/quickbooks_online/expense/operations/upsert.rb
jozefvaclavik: I'm not fan of this, but we need to somehow get operations from final operation to current operation..

ryanwjackson:
@jozefvaclavik Now I think I understand.  I agree that we should be calling `Create`, `Update, `Null`, or whatever and let them handle whatever operations they need.  The `Upsert` should not have to have any knowledge of what takes place in an operation.  It really is a dump operation that just determines whether to do an update or create.  Frankly, we probably could just make it `QBO::Operations::Upsert` and inherit most of the code, the only difference being the resource class.

Issue #31, opened 2019-09-22
Remove sync class


ryanwjackson:
**Is your feature request related to a problem? Please describe.**
The Sync class is a helper method that loosely wraps the functionality of an operation.  It helps with the composition of resources from a structured hash of data, adaptor, etc.  But in return it adds more complexity, such as `SyncResult` and that the utilities are buried by the `Sync` class.

**Describe the solution you'd like**
- Remove the `Sync` class and enable developers to build an operation themselves.
- Determine when/how to call the coordinator.  Perhaps this is just called from the `Performer` directly.  So flow would be to create an adaptor, create your resources (using the `Util::ResourcesBuilder` if desired), and creating the root `Operation`.  Then you would just use the `Performer` to call handle the `Operation`.

**Describe alternatives you've considered**
- Keeping the `Sync` class

**Additional context**
Should we make Sync accept `resources` and `resource` instead of `resources_data` and just have them use `Util::ResourcesBuilder`?

Should we rename "adaptor" to "ledger"?  They are ledgers, but does an adaptor imply the authentication attributes as well?

Is the `Sync` class necessary?

- `.operations` does give you the ordered list
- `SyncResult` does not add much value, and we could get rid of it.
- We could add `:resources` to the `Operation` initializer and it could handle building child operations by passing that down.

- coordinator would need to move.

@mattgmarcus @jozefvaclavik Would love your feedback.

mattgmarcus:
I know less about the inner workings of this, so i trust you if this is a better solution.
Do you have a sense for how hard it'll be to refactor to this? 

Also, on `Remove the Sync class and enable developers to build an operation themselves`

ryanwjackson:
@mattgmarcus it's not hard to refactor really.  The `Sync` class just takes structured primitives and creates an `Adaptor`, `Resources`, and `Operations`.  It does this using utilities, so really it's a wrapper for essentially ~10 lines of code.

Having the `Sync` class also means we need a special type of result, though perhaps the solution here is just to return an `OperationResult` and throw away the `SyncResult` classes.

My thought is that it's actually somewhat harder for a developer as-is because you have to explain how to structure data in primitives where I think that it makes more sense to say:

1. Create an `Adaptor` (which has it's own set of rules based on the ledger)
2. Create your resources (you can use the same utility `Util::ResourcesBuilder` and pass the structured primitive data or you can just build them yourself).
3. Create your operation(s) (which is just passing the adaptor and resource(s))

[...]

PR #52, opened 2019-10-09
Remove Sync class and update README

ryanwjackson:
Fix https://github.com/LedgerSync/ledger_sync/issues/31


README.md
```diff
@@ -22,23 +22,40 @@ Or install it yourself as:
...
 
 ```ruby
-sync = LedgerSync::Sync.new(
-  adaptor: adaptor, # The adaptor for the ledger you want to use (see below)
-  method: :upsert,
-  resources_data: resources_data, # The structured hash of resource data (see below)
-  resource_external_id: resource_external_id, # The root object ID on which the operation is to be performed
-  resource_type: resource_type # The root object type on which the operation is to be performed
+adaptor = LedgerSync::Adaptors::QuickBooksOnline::Adaptor.new(
+  access_token: ENV['QUICKBOOKS_ONLINE_ACCESS_TOKEN'],
```
mattgmarcus: This probably wouldn't be an ENV variable, since it can change right? Just thinking in terms of the example. You could do a line like
```
access_token: access_token # assuming it's defined somewhere above this snippet
```

mattgmarcus: Same for refresh token below

README.md
```diff
@@ -164,24 +181,17 @@ The serialization of any object follows the same struction.  There is a `:root`
...
-    customer: {
-      'cus_1': {
-        name: 'Jane Doe'
-      }
-    }
-  }
-}
-sync = LedgerSync::Sync.new(
+
+sync = LedgerSync::Adaptors::Test::Customer::Operations::Create.new(
```
mattgmarcus: Perhaps name this var `operation` not `sync` to be consistent

README.md
```diff
@@ -22,23 +22,40 @@ Or install it yourself as:
...
 
-sync.operations # Returns the ordered list of operations to be performed
+resource = LedgerSync::Adaptors::QuickBooksOnline::Customer.new(
+  name: 'Sample Customer',
+  email: 'test@example.com'
+)
 
-sync.perform # Returns a SyncResult (either ::Success or ::Failure)
-```
+operation = LedgerSync::Adaptors::QuickBooksOnline::Customer::Create.new(
```
mattgmarcus: Should this include `Operations::` before `Create`? I noticed that below

README.md
```diff
@@ -22,23 +22,40 @@ Or install it yourself as:
...
+)
+
+result = operation.perform # This will be a LedgerSync::OperationResult object
 
-**NOTE: Even when not performing the sync (`sync.perform`) and just retrieving the operations (`sync.operations`), the library will make API requests to the ledger in order to determine if a `create` or `update` is necessary for each object involved.**
+if result.success?
+  # Do something with result.operation.resource
+else # result.failure?
+  raise result.error
+end
```
mattgmarcus: oh @ryanwjackson can you add an example for how to check the latest access/refresh token after doing the operation as well?

PR #53, opened 2019-10-09
Rmeove sub operations from operations

### LINKED ISSUE #47: Remove sub operation calls
After a discussion with @mattgmarcus, we have decided that operations should not depend on other operations (save the example of required full updates on QBO).  The idea here is that we would do the following:

1. Remove Upserts: This functionality can be replicated by the user of the lib (e.g. perform a `Find` to determine if `Create` or `Update` is necessary).  This logic could also me moved to the `Coordinator` or a similar utility class.
2. Remove "sub operation" calls to `Create` or `Update` from all `Create` operations.  We are doing this first as only `Create` will be used for MVP.
3. Remove "sub operation" calls to `Create` or `Update` from all `Update` operations.

The motivations here are the following:

A. As written, there will be unexpected results.  For example, if someone calls the `Expense::Create` operation directly, it currently will attempt to call `Vendor::Upsert`.  There is no way to error if the supplied vendor resource no longer exists in the ledger (which is a use case).
B. Reduce complexity (e.g. needing to think through/handle something like #44).

ryanwjackson:
Fix #47 

This depends on #52 being merged first and the base being changed to `master`.

### DIFF
--- a/lib/ledger_sync/adaptors/quickbooks_online/operation.rb
+++ b/lib/ledger_sync/adaptors/quickbooks_online/operation.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+
+module LedgerSync
+  module Adaptors
+    module QuickBooksOnline
+      module Operation
+        TYPES = %i[create find update upsert].freeze
+
+        module Mixin
+          def self.included(base)
+            base.include Adaptors::Operation::Mixin
+          end
+
+          def perform
+            super
+          rescue OAuth2::Error => e
+            failure(e)
+          end
+        end
+
+        TYPES.each do |type|
+          klass = Class.new do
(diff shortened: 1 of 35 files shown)


lib/ledger_sync/adaptors/quickbooks_online/operation.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+module LedgerSync
+  module Adaptors
+    module QuickBooksOnline
+      module Operation
+        TYPES = %i[create find update upsert].freeze
```
mattgmarcus: We're removing upsert aren't we?

ryanwjackson: Not yet.

PR #62, opened 2019-10-11
Add Deposit resource and Operations

### LINKED ISSUE #22: Add resource support for Deposit (3 hrs)
https://developer.intuit.com/app/developer/qbo/docs/api/accounting/all-entities/deposit

We need to be able to create, update, read, and delete deposits, including using Searchers for this

- ID 
- DepositToAccountRef
- Line
- CurrencyRef
- PrivateNote
- ExchangeRate
- TxnDate

jozefvaclavik:
Fix #22

### DIFF
--- a/lib/ledger_sync/adaptors/quickbooks_online/deposit/operations/create.rb
+++ b/lib/ledger_sync/adaptors/quickbooks_online/deposit/operations/create.rb
@@ -1,0 +1,61 @@
+module LedgerSync
+  module Adaptors
+    module QuickBooksOnline
+      module Deposit
+        module Operations
+          class Create < Operation::Create
+            class Contract < LedgerSync::Adaptors::Contract
+              params do
+                required(:ledger_id).value(:nil)
+                required(:to_account).hash(Types::Reference)
+                required(:currency).filled(:string)
+                required(:memo).filled(:string)
+                required(:transaction_date).filled(:date?)
+                required(:exchange_rate).maybe(:float)
+                required(:line_items).array(Types::Reference)
+              end
+            end
+
+            private
+
+            def operate
+              response = adaptor.post(
+                resource: 'deposit',
+                payload: local_resource_data
+              )
+
+              resource.ledger_id = response.dig('Id')
+              success(response: response)
+            end
+
+            def local_resource_data
+              {
+                'CurrencyRef' => {
+                  'value' => resource.currency,
+                },
+                'TxnDate' => resource.transaction_date.to_s, # Format: YYYY-MM-DD
+                'PrivateNote' => resource.memo,
+                'ExchangeRate' => resource.exchange_rate,
+                'DepositToAccountRef' => {
+                  'value' => resource.to_account.ledger_id
+                },
+                'Line' => resource.line_items.map do |line_item|
+                  {
+                    'DetailType' => 'DepositLineDetail',
+                    'DepositLineDetail' => {
+                      'AccountRef' => {
+                        'value' => line_item.account&.ledger_id || resource.to_account.ledger_id
+                      }
+                    },
+                    'Amount' => line_item.amount / 100.0,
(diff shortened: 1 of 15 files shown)


lib/ledger_sync/adaptors/quickbooks_online/deposit/operations/create.rb
```diff
@@ -0,0 +1,61 @@
...
+                'Line' => resource.line_items.map do |line_item|
+                  {
+                    'DetailType' => 'DepositLineDetail',
+                    'DepositLineDetail' => {
+                      'AccountRef' => {
+                        'value' => line_item.account&.ledger_id || resource.to_account.ledger_id
+                      }
+                    },
+                    'Amount' => line_item.amount / 100.0,
+                    # 'Description' => line_item.description
```
mattgmarcus: @jozefvaclavik why is this commented out

lib/ledger_sync/adaptors/quickbooks_online/deposit/searcher.rb
```diff
@@ -0,0 +1,29 @@
...
+              adaptor
+                .query(
+                  resource: 'deposit',
+                  limit: limit,
+                  offset: offset
+                )
+                .map do |c|
+                  LedgerSync::Deposit.new(
+                    ledger_id: c.fetch('Id'),
+                    name: c.dig('Id')
```
mattgmarcus: Is this a typo? Or are you just saying the ID is the name?

lib/ledger_sync/adaptors/quickbooks_online/deposit/operations/find.rb
```diff
@@ -0,0 +1,36 @@
...
+          class Find < Operation::Find
+            class Contract < LedgerSync::Adaptors::Contract
+              schema do
+                required(:ledger_id).filled(:string)
+                required(:to_account).hash(Types::Reference)
+                required(:currency).maybe(:string)
+                required(:memo).maybe(:string)
+                required(:transaction_date).filled(:date?)
+                required(:exchange_rate).maybe(:float)
+                required(:line_items).array(Types::Reference)
```
mattgmarcus: @jozefvaclavik - it looks like there's a `TotalAmt` field in the API that is read only. Can we maybe fetch that and expose it as an attribute you can read? There's no way to write to it, so we wouldn't let you set it in the create operation

ryanwjackson: @mattgmarcus I'm not sure it's that easy.  The best way to get that info is  `result.operation.response` if it is not an attribute of the resource.  In my serializer PR, I am making it so it merges in any other values on create/update/find as opposed to just setting the ID or even leaving it as what we passed in.

lib/ledger_sync/adaptors/quickbooks_online/deposit/searcher.rb
jozefvaclavik: Actually in this case it doesnt make sense to set name. As thats not an attribute. Name is computed property.

lib/ledger_sync/adaptors/quickbooks_online/deposit/operations/find.rb
jozefvaclavik: Yea, once we have serialization, we can do that.

lib/ledger_sync/adaptors/quickbooks_online/deposit/operations/create.rb
jozefvaclavik: Ups, sigh.

lib/ledger_sync/adaptors/quickbooks_online/deposit/operations/find.rb
mattgmarcus: Kk sounds good

Issue #48, opened 2019-10-08
Add x_refresh_token_expires_in to QBO attributes_to_save


ryanwjackson:
- [ ] Ensure we are saving all the updated parameters (including `x_refresh_token_expires_in`): https://developer.intuit.com/app/developer/qbo/docs/develop/authentication-and-authorization/oauth-2.0
- [ ] Document `attributes_to_save` and psuedocode for when to save.

mattgmarcus:
@ryanwjackson  I noticed that the `expires_at` field is often `nil`. Just wanted to double check that it isn't a typo

PR #64, opened 2019-10-11
Add refresh_token_expires_at to adaptor and attributes to save

ryanwjackson:
Fix https://github.com/LedgerSync/ledger_sync/issues/48

### DIFF
--- a/lib/ledger_sync/adaptors/quickbooks_online/adaptor.rb
+++ b/lib/ledger_sync/adaptors/quickbooks_online/adaptor.rb
@@ -81,7 +81,7 @@ module LedgerSync
           @access_token = refreshed.token
 
           @expires_at = Time&.at(refreshed.expires_at.to_i)&.to_datetime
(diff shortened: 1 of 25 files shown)


lib/ledger_sync/adaptors/quickbooks_online/adaptor.rb
```diff
@@ -80,6 +81,7 @@ def refresh!
           @access_token = refreshed.token
 
           @expires_at = Time&.at(refreshed.expires_at.to_i)&.to_datetime
+          @refresh_token_expires_at = Time&.at(Time.now.to_i + refreshed.params['x_refresh_token_expires_in'])&.to_datetime
```
mattgmarcus: You should make this nilsafe if `refreshed.params['x_refresh_token_expires_in']` is `nil`

ryanwjackson: Done, but it's in the spec so it really shouldn't be nil or we probably have other problems.

PR #68, opened 2019-10-13
Add Bill resource and operations

### LINKED ISSUE #28: Add resource support for Bill (4 hrs)
https://developer.intuit.com/app/developer/qbo/docs/api/accounting/all-entities/bill

We need to be able to create, update, read, and delete bills, including using Searchers for this

- Id
- VendorRef
- Line (AccountBasedExpenseLine)
- CurrencyRef
- TxnDate
- APAccountRef
- DueDate
- PrivateNote

jozefvaclavik:
Fix #28

### DIFF
--- a/lib/ledger_sync/adaptors/quickbooks_online/bill/operations/create.rb
+++ b/lib/ledger_sync/adaptors/quickbooks_online/bill/operations/create.rb
@@ -1,0 +1,65 @@
+module LedgerSync
+  module Adaptors
+    module QuickBooksOnline
+      module Bill
+        module Operations
+          class Create < Operation::Create
+            class Contract < LedgerSync::Adaptors::Contract
+              params do
+                required(:ledger_id).value(:nil)
+                required(:vendor).hash(Types::Reference)
+                required(:account).hash(Types::Reference)
+                required(:currency).filled(:string)
+                required(:memo).maybe(:string)
+                required(:transaction_date).maybe(:date?)
+                required(:due_date).maybe(:date?)
+                required(:line_items).array(Types::Reference)
+              end
+            end
+
+            private
+
+            def operate
+              response = adaptor.post(
+                resource: 'bill',
+                payload: local_resource_data
+              )
+
+              resource.ledger_id = response.dig('Id')
+              success(response: response)
+            end
+
+            def local_resource_data
+              {
+                'CurrencyRef' => {
+                  'value' => resource.currency,
+                },
+                'TxnDate' => resource.transaction_date.to_s, # Format: YYYY-MM-DD
+                'DueDate' => resource.due_date.to_s, # Format: YYYY-MM-DD
+                'PrivateNote' => resource.memo,
+                'VendorRef' => {
+                  'value' => resource.vendor.ledger_id,
+                },
+                'APAccountRef' => {
+                  'value' => resource.account.ledger_id
+                },
+                'Line' => resource.line_items.map do |line_item|
+                  {
+                    'DetailType' => 'AccountBasedExpenseLineDetail',
+                    'AccountBasedExpenseLineDetail' => {
+                      'AccountRef' => {
+                        'value' => line_item.account&.ledger_id
+                      }
+                    },
+                    'Amount' => line_item.amount / 100.0,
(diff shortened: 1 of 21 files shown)


lib/ledger_sync/adaptors/quickbooks_online/bill/operations/create.rb
```diff
@@ -0,0 +1,65 @@
...
+                'PrivateNote' => resource.memo,
+                'VendorRef' => {
+                  'value' => resource.vendor.ledger_id,
+                },
+                'APAccountRef' => {
+                  'value' => resource.account.ledger_id
+                },
+                'Line' => resource.line_items.map do |line_item|
+                  {
+                    'DetailType' => 'AccountBasedExpenseLineDetail',
```
jozefvaclavik: Yeah, Bill is using ExpenseLineDetail. I thought about reusing our existing line item, but different ledger could be using different structure. So let it be separate `BillLineItem`

mattgmarcus: Yeah I think this makes sense

lib/ledger_sync/adaptors/quickbooks_online/bill/operations/create.rb
```diff
@@ -0,0 +1,65 @@
...
+                    'AccountBasedExpenseLineDetail' => {
+                      'AccountRef' => {
+                        'value' => line_item.account&.ledger_id || resource.account.ledger_id
```
mattgmarcus: Can you see if this would work in practice? I saw that the APAccountRef has to be a liability account whereas the line item account has to be an expense one. So it's possible that this fallback would fail, in which case you should have this just be `'value' => line_item.account&.ledger_id`
Unfortunately you're going to have to pull down some specific accounts from your testing environment to try that out

lib/ledger_sync/adaptors/quickbooks_online/bill/operations/create.rb
```diff
@@ -0,0 +1,65 @@
...
+  module Adaptors
+    module QuickBooksOnline
+      module Bill
+        module Operations
+          class Create < Operation::Create
+            class Contract < LedgerSync::Adaptors::Contract
+              params do
+                required(:ledger_id).value(:nil)
+                required(:vendor).hash(Types::Reference)
+                required(:account).hash(Types::Reference)
```
mattgmarcus: I think this field is optional no? As are due date, transaction date, and due memo

lib/ledger_sync/adaptors/quickbooks_online/bill/operations/create.rb
jozefvaclavik: Yeah, they needs to be different. 👌

mattgmarcus: Also we may have the same issue for Expenses or other places we have this type of fallback logic. Can you double check our other resources?

lib/ledger_sync/adaptors/quickbooks_online/bill/operations/create.rb
jozefvaclavik: Added dates and memo as optional, but relationship is _optional_ by default. Our schema doesn't validate required relationship atm.

lib/ledger_sync/adaptors/quickbooks_online/bill/operations/create.rb
jozefvaclavik: Yeah, this issue showed up now during Bill. Just forgot to fix this line. Thats why I left a note in spec about `account1` being specific type and different then `account2`.

For the Deposit they doesnt need to be necessarily different, but account on line item needs to be `Income` type.

And similar for Expense, line item accounts needs to be `Expense` type.

I'll remove fallback completely.

lib/ledger_sync/adaptors/quickbooks_online/bill/operations/create.rb
mattgmarcus: Ah ok...well should we just make it optional anyway so that when (If?) it is required, this won't fail?

PR #70, opened 2019-10-13
Add DisplayName to Vendor

### LINKED ISSUE #69: Add display_name attribute to vendor
There's another attribute called display_name that you can set on the vendor. We'd use this for companies, whereas we'd use first and last name for individuals. I'd like support for this for our operations/searcher

jozefvaclavik:
Fix #69


lib/ledger_sync/adaptors/quickbooks_online/vendor/operations/update.rb
```diff
@@ -34,6 +35,7 @@ def operate
 
             def local_resource_data
               {
+                'Displayname' => resource.display_name,
```
mattgmarcus: Looks like the `n` isn't capitalized

PR #72, opened 2019-10-13
Add JournalEntry resource and operations

### LINKED ISSUE #27: Add resource support for JournalEntry (5 hrs)
https://developer.intuit.com/app/developer/qbo/docs/api/accounting/all-entities/journalentry

We need to be able to create, update, read, and delete journal entries, including using Searchers for this

- Id
- Line
- CurrencyRef
- PrivateNote
- TxnDate

jozefvaclavik:
Fix #27

### DIFF
--- a/lib/ledger_sync/adaptors/quickbooks_online/adaptor.rb
+++ b/lib/ledger_sync/adaptors/quickbooks_online/adaptor.rb
@@ -44,10 +44,10 @@ module LedgerSync
         end
 
         def find(resource:, id:)
-          url = "#{oauth_base_uri}/#{underscore(resource)}/#{id}"
+          url = "#{oauth_base_uri}/#{resourcify(resource)}/#{id}"
 
           response = request(:get, url, headers: OAUTH_HEADERS.dup)
-          JSON.parse(response.body).dig(resource.capitalize)
+          JSON.parse(response.body).dig(resource.classify)
         end
 
         def parse_operation_error(error:, operation:)
@@ -60,18 +60,18 @@ module LedgerSync
         end
 
         def post(resource:, payload:)
-          url = "#{oauth_base_uri}/#{underscore(resource)}"
+          url = "#{oauth_base_uri}/#{resourcify(resource)}"
 
           response = request(:post, url, headers: OAUTH_HEADERS.dup, body: payload.to_json)
-          JSON.parse(response.body).dig(resource.capitalize)
+          JSON.parse(response.body).dig(resource.classify)
         end
 
         def query(resource:, query:, limit: 10, offset: 1)
-          full_query = "SELECT * FROM #{resource.capitalize} WHERE #{query} STARTPOSITION #{offset} MAXRESULTS #{limit}"
+          full_query = "SELECT * FROM #{resource.classify} WHERE #{query} STARTPOSITION #{offset} MAXRESULTS #{limit}"
           url = "#{oauth_base_uri}/query?query=#{CGI.escape(full_query)}"
 
           response = request(:get, url, headers: OAUTH_HEADERS.dup)
-          JSON.parse(response.body).dig('QueryResponse', resource.capitalize) || []
+          JSON.parse(response.body).dig('QueryResponse', resource.classify) || []
         end
 
         def refresh!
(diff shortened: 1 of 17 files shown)


lib/ledger_sync/adaptors/quickbooks_online/adaptor.rb
```diff
@@ -44,10 +44,10 @@ def initialize(
         end
 
         def find(resource:, id:)
-          url = "#{oauth_base_uri}/#{underscore(resource)}/#{id}"
+          url = "#{oauth_base_uri}/#{resourcify(resource)}/#{id}"
```
ryanwjackson: Why this change instead of using `Util::StringHelpers`?

jozefvaclavik: its just `.tr('_', '')`. Didn't feel to me that it needs to be as Util. Should I move it as utility helper?

ryanwjackson:
I believe we use it other places is why I put it in a Util.


--
Ryan Jackson
ryanwjackson@gmail.com

lib/ledger_sync/adaptors/quickbooks_online/adaptor.rb
mattgmarcus: This might be QBO specific though. I'm fine with it like this

lib/ledger_sync/adaptors/quickbooks_online/journal_entry/operations/create.rb
```diff
@@ -0,0 +1,56 @@
...
+                },
+                'TxnDate' => resource.transaction_date.to_s, # Format: YYYY-MM-DD
+                'PrivateNote' => resource.memo,
+                'Line' => resource.line_items.map do |line_item|
+                  {
+                    'DetailType' => 'JournalEntryLineDetail',
+                    'JournalEntryLineDetail' => {
+                      'PostingType': Mapping::LINE_ITEM_TYPES[line_item.entry_type],
+                      'AccountRef' => {
+                        'value' => line_item.account&.ledger_id
```
mattgmarcus: Why make this nil safe?

lib/ledger_sync/adaptors/quickbooks_online/journal_entry/operations/update.rb
```diff
@@ -0,0 +1,60 @@
...
+                },
+                'TxnDate' => resource.transaction_date.to_s, # Format: YYYY-MM-DD
+                'PrivateNote' => resource.memo,
+                'Line' => resource.line_items.map do |line_item|
+                  {
+                    'DetailType' => 'JournalEntryLineDetail',
+                    'JournalEntryLineDetail' => {
+                      'PostingType': Mapping::LINE_ITEM_TYPES[line_item.entry_type],
+                      'AccountRef' => {
+                        'value' => line_item.account&.ledger_id
```
mattgmarcus: Same Q about nil safe

lib/ledger_sync/adaptors/quickbooks_online/journal_entry/searcher.rb
```diff
@@ -0,0 +1,28 @@
...
+            @resources ||= begin
+              adaptor
+                .query(
+                  resource: 'journal_entry',
+                  limit: limit,
+                  offset: offset
+                )
+                .map do |c|
+                  LedgerSync::JournalEntry.new(
+                    ledger_id: c.fetch('Id')
```
mattgmarcus: Can we pull in the other attributes while you're here?

lib/ledger_sync/adaptors/quickbooks_online/journal_entry/operations/create.rb
jozefvaclavik: In case we don't have account, right now we will let QBO fail, instead of crashing in the lib. This will change once we handle serialisation and dirty.

Right now we would send `nil` to QBO, which could remove account if its optional attribute. With serialisation we will not serialise it if we don't provide account.

lib/ledger_sync/adaptors/quickbooks_online/journal_entry/searcher.rb
jozefvaclavik: Yes, but this will get automated once we do serialisation.

PR #74, opened 2019-10-15
Deep merge hashes in array

jozefvaclavik:
_Some_ line items has IDs. The way we merged payloads was to override arrays in fetched payload with values from our payload.

QBO luckily handles update as _replace_. So any previous line items gets deleted and new are created.

The issue becomes if user changes some attribute in QBO on a line item (lets say tax category) that we don't handle.

If we don't preserve ID of line item and push update, we basically remove that line item and create new one with _default_ values for attributes we don't set (mostly null).

If we preserve ID of line item and we pass it as it in next update, QBO performs update on a line item instead of creating new one.

This PR updates `merge_into` method to deep merge arrays of hashes. It does `map` on new array and tries to lookup/find match in old array by `Id` (which is QBO specific) and merges new item into old item.

For example:
QBO responds: 
```
{
  Line: [
    {Id: 1, Description: 'This', Amount: 10},
    {Id: 2, Description: 'Or this', Amount: 20},
  ]
}
```

and our update is:
```
{
  Line: [
    {Id: 1, Description: 'Noooo'}
  ]
}
```

then resulting payload will be:
```
{
  Line: [
    {Id: 1, Description: 'Noooo', Amount: 10}
  ]
}
```

This PR doesnt handle arrays of arrays with hashes `[ [{}], [{}, {}] ]`, if we will need it, we can do that later.

---
Note 1: `merge_into` is being moved to QBO as its ledger specific in Ryans serializer PR.
Note 2: Some line items starts with ID: 1 and some starts with ID: 0. Just sayin.

### DIFF
--- a/lib/ledger_sync/adaptors/operation.rb
+++ b/lib/ledger_sync/adaptors/operation.rb
@@ -178,12 +178,27 @@ module LedgerSync
 
         def merge_into(from:, to:)
           case to
-          when *(Resource::PRIMITIVES | [Array]) then from
+          when *Resource::PRIMITIVES then from
+          when Array then merge_list(from: from, to: to)
           else to.merge!(from) { |_key, new_value, old_value| merge_into(from: old_value, to: new_value) } if to && from
           end
         end
 
         private
+
+        def merge_list(from:, to:)
+          from.map do |from_item|
+            if from_item.is_a?(Hash) && (match = lookup_match(item: from_item, list: to))
+              match.merge!(from_item) { |_key, new_value, old_value| merge_into(from: old_value, to: new_value) }
+            else
+              from_item
+            end
+          end
+        end
+
+        def lookup_match(item:, list:)
+          list.find { |li| li.fetch('Id', false) == item['Id'] }
+        end
 
         def operate
           raise NotImplementedError, self.class.name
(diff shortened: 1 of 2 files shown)


lib/ledger_sync/adaptors/operation.rb
```diff
@@ -178,13 +178,28 @@ def ==(other)
...
+            if from_item.is_a?(Hash) && (match = lookup_match(item: from_item, list: to))
+              match.merge!(from_item) { |_key, new_value, old_value| merge_into(from: old_value, to: new_value) }
+            else
+              from_item
+            end
+          end
+        end
+
+        def lookup_match(item:, list:)
+          list.find { |li| li.fetch('Id', false) == item['Id'] }
```
jozefvaclavik: This feels bit like a hack. I just didn't wanna write it as: `li['Id'] && li['Id'] == item['Id']`, but I'm fine with it either way.

ryanwjackson:
@jozefvaclavik I am closing this as I believe it is old.  Please re-open if I'm wrong.

ryanwjackson:
https://github.com/LedgerSync/ledger_sync/blob/master/spec/adaptors/quickbooks_online/ledger_serializer_spec.rb#L136

PR #84, opened 2019-10-18
Add ledger_id and external_id to resource_attributes

### LINKED ISSUE #83: Add ledger_id and external_id to resource_attributes
You have to do extra work to get a hash of all resource attributes. 

Be able to call `resource.to_h` or `resource.resource_attributes.to_h` and receive all resource-specific attributes as well as `ledger_id` and `external_id` in a hash.

Something like: 

```ruby
{
  ledger_id: '',
  external_id: '',
  attributes: {
    name: 'asdf'
  }
}
```

This was in an attempt to avoid conflicts of the resource object attributes (e.g. `ledger_id`) and resource type-specific attributes (e.g. Customer has a `name` attribute).  But it seems eventually, you will have to handle conflicts.

ryanwjackson:
Fix #83

### DIFF
--- a/lib/ledger_sync/adaptors/test/account/operations/valid.rb
+++ b/lib/ledger_sync/adaptors/test/account/operations/valid.rb
@@ -7,7 +7,6 @@ module LedgerSync
             class Contract < LedgerSync::Adaptors::Contract
               schema do
                 optional(:external_id).maybe(:string)
-                optional(:external_id).maybe(:string)
                 optional(:ledger_id).maybe(:string)
                 required(:name).filled(:string)
                 required(:classification).filled(:string)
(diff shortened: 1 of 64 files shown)


lib/ledger_sync/adaptors/test/account/operations/valid.rb
```diff
@@ -6,6 +6,8 @@ module Operations
           class Valid < Operation::Create
             class Contract < LedgerSync::Adaptors::Contract
               schema do
+                optional(:external_id).maybe(:string)
+                optional(:external_id).maybe(:string)
```
jozefvaclavik: 2x

lib/ledger_sync/resource.rb
```diff
@@ -65,13 +56,17 @@ def to_h
       resource_attributes.to_h.merge(dirty_attributes_to_h)
     end
 
+    def self.inherited(subclass)
```
jozefvaclavik: Oh, nice. I was wondering how would handle that.

PR #93, opened 2019-10-24
Turns out that the production URL should not be prefixed with app

mattgmarcus:
Your domain is unique and the redirect is not preserved


ryanwjackson:
@mattgmarcus Do we want to publish this to 1.1.3 and then bump the current milestone to 1.1.4?

mattgmarcus:
Maybe we should just divorce the milestone number from the version number, so we don't have to keep bumping it.

Let's merge https://github.com/LedgerSync/ledger_sync/pull/94 too first

PR #94, opened 2019-10-25
Allow polymorphic entity on expense

### LINKED ISSUE #92: Let you pass Customers as the EntityRef to Expense
We currently only allow you to pass in a vendor to the expense object. QBO also lets you do customers or employees:
https://developer.intuit.com/app/developer/qbo/docs/api/accounting/all-entities/purchase
image

Since we only have customer right now, let's add in support for that. 
I'm not sure how you want to do this, since now it can be either the vendor or the customer that is present. Perhaps we need some concept of a polymorphic entity?

cc @ryanwjackson

ryanwjackson:
Fix #92


mattgmarcus:
@ryanwjackson in addition to the errors I mentioned in Slack, can we see if we can get this to work for the DepositLineItem resource too. There's an entity field on it, as described here:
https://developer.intuit.com/app/developer/qbo/docs/api/accounting/all-entities/deposit

image

The syntax of it seems slightly different so I just want to make sure this format extends to it okay. Plus, I realized I need that too so this is my way of adding it into the current milestone :)

Issue #66, opened 2019-10-11
Utilities for processing webhooks


mattgmarcus:
**Is your feature request related to a problem? Please describe.**
QBO can send you webhooks. I'd like some helper functions to help out with processing these.
The webhooks are generally pretty unspecific. It just gives you a body like this
```
{
   "eventNotifications":[
   {
      "realmId":"1185883450",
      "dataChangeEvent":
      {
         "entities":[
         {
            "name":"Customer",
            "id":"1",
            "operation":"Create",
            "lastUpdated":"2015-10-05T14:42:19-0700"
         },
         {
            "name":"Vendor",
            "id":"1",
            "operation":"Create",
            "lastUpdated":"2015-10-05T14:42:19-0700"
         }]
      }
   }]
}
```
https://developer.intuit.com/app/developer/qbo/docs/develop/webhooks/managing-webhooks-notifications

They also include a token you can use to verify that it's from them

**Describe the solution you'd like**
I need a way to verify the request came from QBO. I expect the method signature is something like
```
def verify_quickbooks_token(verifier_token, request_body) --> returns true/false
```
I'm not sure what you pass in exactly. Probably the whole body.

**Describe alternatives you've considered**
Do we want a lightweight function that just translates the body into ledgersync resources
```
def parse_webhook(webhook_body)

--> returns
{
  realm_id: realm_id,
  entities: [array of ledgersync entities with their ledger_id],
}
```

I'm not sure if that would really be helpful. Plus we still have to capture the event action (create, delete, etc).

We could add a way to do a `Find` on the entity automatically but you could do that yourself too as the consumer.

Originally I was leaning towards doing this but not I'm not. Thoughts @ryanwjackson ?

ryanwjackson:
@mattgmarcus We can handle the verification, but we could also add `QuickBooksOnline::Webhook` and `QuickBooksOnline::WebhookEvent` class that can do both the verification and parsing.  With that you could get something like `webhook.events.first.operation` and probably event `webhook.events.first.resource(s)`.

Thoughts?

mattgmarcus:
Yeah I think that would be nice. What would `.operation` return though?

ryanwjackson:
Oh, good question.  Not our operation.  I was just reading their key.  We could call it `quickbooks_method`.  I'm basing this solely on the sample webhook you sent.  Probably worth looking at a few more to make sure they all follow that structure.

PR #95, opened 2019-11-01
Adds webhook parsing for QBO

ryanwjackson:
Fixes https://github.com/LedgerSync/ledger_sync/issues/66

See the updated readme for how to use.  Let me know if it isn't clear.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,40 @@ previous_searcher = searcher.previous_searcher
 
 ## QuickBooks Online
 
+### Webhooks
+
+Reference: https://developer.intuit.com/app/developer/qbo/docs/develop/webhooks/managing-webhooks-notifications#validating-the-notification
+
+LedgerSync offers an easy way to validate and parse webhook payloads.  It also allows you to easily fetch the resources referenced.  You can create and use a webhook with the following:
+
+```ruby
+# Assuming `request` is the webhook request received from Quickbooks Online
+webhook = LedgerSync::Adaptors::QuickBooksOnline::Webhook.new(
+  payload: request.body.read # It accepts a JSON string or hash
+)
+
+verification_token = WEBHOOK_VERIFICATION_TOKEN # You get this token when you create webhooks in the QuickBooks Online dashboard
+signature = request.headers['intuit-signature']
+raise 'Not valid' unless webhook.valid?(signature: signature, verification_token: verification_token)
+
+# Although not yet used, webhooks may include notifications for multiple realms
+webhook.notifications.each do |notification|
+  puts notification.realm_id
+
+  # Multiple events may be referenced.
+  notification.events.each do |event|
+    puts event.resource # This will perform a Find operation for the resource retrieving the latest version from QuickBooks Online
+  end
+
+  # Other helpful methods
+  notification.resources # All events for a given webhook across all events
+end
+
+# Other helpful methods
+webhook.events # All events for a given webhook across all realms
+webhook.resources # All events for a given webhook across all realms and events
+```
+
 ### Errors
 
 https://developer.intuit.com/app/developer/qbo/docs/develop/troubleshooting/error-codes
--- a/README.md
+++ b/README.md
@@ -162,11 +162,16 @@ webhook.notifications.each do |notification|
 
   # Multiple events may be referenced.
   notification.events.each do |event|
-    puts event.resource # This will perform a Find operation for the resource retrieving the latest version from QuickBooks Online
+    puts event.resource # Returns a LedgerSync resource with the `ledger_id` set
+
+    # Other helpful methods
+    notification.find_operation_class(adaptor: your_quickbooks_adaptor_instance) # The respective Find class
+    notification.find_operation(adaptor: your_quickbooks_adaptor_instance) # The initialized respective Find operation
+    notification.find(adaptor: your_quickbooks_adaptor_instance) # Performs a Find operation for the resource retrieving the latest version from QuickBooks Online
   end
 
   # Other helpful methods
-  notification.resources # All events for a given webhook across all events
+  notification.resources # All resources for a given webhook across all events
 end
 
 # Other helpful methods
--- a/lib/ledger_sync/adaptors/quickbooks_online/webhook.rb
+++ b/lib/ledger_sync/adaptors/quickbooks_online/webhook.rb
@@ -1,0 +1,59 @@
+# frozen_string_literal: true
+
+require_relative 'webhook_event'
+
+# ref: https://developer.intuit.com/app/developer/qbo/docs/develop/webhooks/managing-webhooks-notifications#validating-the-notification
+module LedgerSync
+  module Adaptors
+    module QuickBooksOnline
+      class Webhook
+        attr_reader :notifications,
+                    :original_payload,
+                    :payload
+
+        def initialize(payload:)
+          @original_payload = payload
+          payload = JSON.parse(payload) if payload.is_a?(String)
+          @payload = payload
+
+          event_notifications_payload = payload.dig('eventNotifications')
+          raise 'Invalid payload: Could not find eventNotifications' if event_notifications_payload.blank?
+
+          @notifications = []
+
+          event_notifications_payload.each do |event_notification_payload|
+            @notifications << WebhookNotification.new(
+              payload: event_notification_payload,
+              webhook: self
+            )
+          end
+        end
+
+        def events
+          notifications.map(&:events)
+        end
+
+        def resources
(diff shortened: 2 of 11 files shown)


lib/ledger_sync/adaptors/quickbooks_online/webhook.rb
```diff
@@ -0,0 +1,59 @@
...
+  module Adaptors
+    module QuickBooksOnline
+      class Webhook
+        attr_reader :notifications,
+                    :original_payload,
+                    :payload
+
+        def initialize(payload:)
+          @original_payload = payload
+          payload = JSON.parse(payload) if payload.is_a?(String)
```
mattgmarcus: I generally prefer not mutating params passed in. Maybe
```
@payload = payload.is_a?(String) ? JSON.parse(payload) : payload
```

lib/ledger_sync/adaptors/quickbooks_online/webhook.rb
```diff
@@ -0,0 +1,59 @@
...
+          @payload = payload
+
+          event_notifications_payload = payload.dig('eventNotifications')
+          raise 'Invalid payload: Could not find eventNotifications' if event_notifications_payload.blank?
```
mattgmarcus: I'd rather not raise an error and just return empty resources/notifications. This seems like risky behavior since QBO is hard to trust

lib/ledger_sync/adaptors/quickbooks_online/webhook_event.rb
```diff
@@ -0,0 +1,71 @@
...
+                    :ledger_id,
+                    :original_payload,
+                    :payload,
+                    :quickbooks_online_resource_type,
+                    :webhook_notification,
+                    :webhook
+
+        def initialize(payload:, webhook_notification: nil)
+          @original_payload = payload
+          payload = JSON.parse(payload) if payload.is_a?(String)
```
mattgmarcus: Same comment as above about handling payload

lib/ledger_sync/adaptors/quickbooks_online/webhook_event.rb
```diff
@@ -0,0 +1,71 @@
...
+          @payload = payload
+
+          @deleted_id = payload.dig('deletedId')
+
+          @event_operation = payload.dig('operation')
+          raise 'Invalid payload: Could not find operation' if @event_operation.blank?
```
mattgmarcus: Same as above about not failing. Would rather return an empty object or nil

lib/ledger_sync/adaptors/quickbooks_online/webhook_event.rb
```diff
@@ -0,0 +1,71 @@
...
+
+          @quickbooks_online_resource_type = payload.dig('name')
+          raise 'Invalid payload: Could not find name' if @quickbooks_online_resource_type.blank?
+
+          @webhook_notification = webhook_notification
+          @webhook = webhook_notification.try(:webhook)
+        end
+
+        def find(adaptor:)
+          adaptor.class.base_operation_module_for(resource_class: resource_class)::Find.new(
```
mattgmarcus: Can you do a check if the Find operation exists, as well as if the base operations exist? I imagine we'll get webhooks for some things we don't have resources for

lib/ledger_sync/adaptors/quickbooks_online/webhook_event.rb
```diff
@@ -0,0 +1,71 @@
...
+            adaptor: adaptor,
```
mattgmarcus: How is the adaptor getting the right credentials do actually do the find operation?

README.md
```diff
@@ -140,6 +140,40 @@ previous_searcher = searcher.previous_searcher
...
+signature = request.headers['intuit-signature']
+raise 'Not valid' unless webhook.valid?(signature: signature, verification_token: verification_token)
+
+# Although not yet used, webhooks may include notifications for multiple realms
+webhook.notifications.each do |notification|
+  puts notification.realm_id
+
+  # Multiple events may be referenced.
+  notification.events.each do |event|
+    puts event.resource # This will perform a Find operation for the resource retrieving the latest version from QuickBooks Online
```
mattgmarcus: Can we expose methods to get the resources without doing finds? In some cases, I feel like we won't have to do that. Like, if an object is deleted. They could be like `raw_resource` and `raw_resources`, and just init the ledger class with the ledger_id

README.md
```diff
@@ -140,6 +140,40 @@ previous_searcher = searcher.previous_searcher
...
+  end
+
+  # Other helpful methods
+  notification.resources # All events for a given webhook across all events
```
mattgmarcus: `All events for a given webhook across all events` should the second word be `resources`?

lib/ledger_sync/adaptors/quickbooks_online/webhook_notification.rb
```diff
@@ -0,0 +1,43 @@
...
+  module Adaptors
+    module QuickBooksOnline
+      class WebhookNotification
+        attr_reader :events,
+                    :original_payload,
+                    :payload,
+                    :realm_id,
+                    :webhook
+
+        def initialize(payload:, webhook: nil)
```
mattgmarcus: Same comment here about payload and errors

lib/ledger_sync/adaptors/quickbooks_online/webhook.rb
ryanwjackson: True.  But I actually just intended for this to check for an array.  It will error out below because it can't parse anything if not.  I figured this error was friendlier than chasing down why it was erroring in the first place.

lib/ledger_sync/adaptors/quickbooks_online/webhook_event.rb
ryanwjackson: True.  How do you want to handle that downstream if it doesn't exist?

lib/ledger_sync/adaptors/quickbooks_online/webhook_event.rb
ryanwjackson: Sorry, I don't understand.

README.md
ryanwjackson: I actually was thinking about that.  Will add it in.

lib/ledger_sync/adaptors/quickbooks_online/webhook_event.rb
mattgmarcus: Hmm. Good question. We could have a class like `UnsupportedResource` and just return that with the ledger_id on it. That way you can iterate past it.
The other way is to return `nil` here and then `compact` (optionally) where we get the events.

ryanwjackson: I actually made some updates.  Let me know what you think.  I think this is clearer, letting `.resource` to return a local resource with just the `ledger_id` set.  Then there are helper methods for performing the finds yourself if you want.

We still have the issue that it will error if `resource_class` does not exist (e.g. an object we haven't created yet).  For now, I have `resource` returning `nil` if that's the case.

README.md
ryanwjackson: See if you agree with https://github.com/LedgerSync/ledger_sync/pull/95#discussion_r341659679 which should resolve this.

ryanwjackson:
@mattgmarcus I'll let you merge this when you get to test it.

lib/ledger_sync/adaptors/quickbooks_online/webhook.rb
```diff
@@ -0,0 +1,58 @@
...
+  module Adaptors
+    module QuickBooksOnline
+      class Webhook
+        attr_reader :notifications,
+                    :original_payload,
+                    :payload
+
+        def initialize(payload:)
+          @original_payload = payload
+          @payload = payload.is_a?(String) ? JSON.parse(payload) : payload
```
mattgmarcus: Only comment I have is that I had to pass in my args as `payload.with_indifferent_access`. It could be nice to do that here instead, so the caller doesn't have to.

mattgmarcus: Bump  @ryanwjackson

ryanwjackson: @mattgmarcus `.with_indifferent_access` is Rails (I believe) and not pure ruby.

mattgmarcus: Ah yes, that's right
Okay so I'll just merge

Issue #82, opened 2019-10-17
Add helper method to revoke token


mattgmarcus:
**Is your feature request related to a problem? Please describe.**
The user has to be able to disconnect from Quickbooks

**Describe the solution you'd like**
I'd like a helper method on the QBO adaptor to revoke the token
It's documented here https://developer.intuit.com/app/developer/qbo/docs/develop/authentication-and-authorization/oauth-2.0#revoke-token-disconnect
![image](https://user-images.githubusercontent.com/4755444/67050010-6214f700-f0ec-11e9-8e8f-5e1829a5e792.png)

PR #96, opened 2019-11-01
Add method to revoke qbo access token

ryanwjackson:
Fixes https://github.com/LedgerSync/ledger_sync/issues/82


lib/ledger_sync/adaptors/quickbooks_online/adaptor.rb
```diff
@@ -96,6 +97,17 @@ def refresh!
           raise parse_error(error: e)
         end
 
+        def revoke_token!
```
mattgmarcus: I expect methods that end in a bang to raise an error if they fail. So I'd say either do that or change to `revoke_token`

mattgmarcus: Ah nvm, that sounds good then

PR #105, opened 2019-12-05
Add Class and Department

### LINKED ISSUE #25: Add resource support for Class (3 hrs)
https://developer.intuit.com/app/developer/qbo/docs/api/accounting/all-entities/class

We need to be able to create, update, read, and delete classes, including using Searchers for this

- Id
- Name
- ParentRef
- SubClass
- Active
- FullyQualifiedName

### LINKED ISSUE #26: Add resource support for Department (2 hrs)
https://developer.intuit.com/app/developer/qbo/docs/api/accounting/all-entities/department

We need to be able to create, update, read, and delete departments, including using Searchers for this

- Id
- Name
- ParentRef
- Active
- FullyQualifiedName
- SubDepartment

jozefvaclavik:
Fix #25 
Fix #26 

Adds `LedgerClass` and `Department` resources and their operations.

Known issues: Cannot create root/parent Department. This _should_ be fixed on our side, its just out of the scope of this PR.

Its OK to send this `"ParentRef":{"value":null}` with Class, but its not OK to send it with Department. This could be just sandbox API issue. Would be best to test this on production as well. Because the error is not meaningful error, its an API crash (`Unexpected Internal Error `).
```
{"Fault":{"Error":[{"Message":"A business validation error has occurred while processing your request","Detail":"Business Validation Error: Unexpected Internal Error. (-30000)","code":"6000","element":""}],"type":"ValidationFault"},"time":"2019-12-04T11:50:11.923-08:00"}>
```


lib/ledger_sync/adaptors/quickbooks_online/department/ledger_serializer.rb
```diff
@@ -0,0 +1,37 @@
...
+                    resource_attribute: :active
+          attribute ledger_attribute: 'SubDepartment',
+                    resource_attribute: :sub_department
+          attribute ledger_attribute: 'FullyQualifiedName',
+                    resource_attribute: :fully_qualified_name
+
+          attribute ledger_attribute: 'ParentRef.value',
+                    resource_attribute: 'parent.ledger_id'
+
+          def to_ledger_hash(deep_merge_unmapped_values: {}, only_changes: false)
```
ryanwjackson: @jozefvaclavik can you add some comments above this explaining why we have to do this?

lib/ledger_sync/adaptors/quickbooks_online/ledger_class/ledger_serializer.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module LedgerSync
+  module Adaptors
+    module QuickBooksOnline
+      module LedgerClass
+        class LedgerSerializer < QuickBooksOnline::LedgerSerializer
+          quickbooks_online_resource_type :class
+
+          id  ledger_attribute: 'Id',
```
ryanwjackson: You can just call `id` as it defaults to these values.

spec/adaptors/quickbooks_online/operations/ledger_class/create_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+support :adaptor_helpers
+
+RSpec.describe LedgerSync::Adaptors::QuickBooksOnline::LedgerClass::Operations::Create do
+  include AdaptorHelpers
+
+  let(:ledger_class) { LedgerSync::LedgerClass.new(name: 'Test', active: true, sub_class: false) }
+
+  it do
+    instance = described_class.new(resource: ledger_class, adaptor: quickbooks_adaptor)
+    instance.valid?
```
ryanwjackson: Remove this line.

spec/adaptors/quickbooks_online/operations/ledger_class/update_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'spec_helper'
+
+support :adaptor_helpers
+
+RSpec.describe LedgerSync::Adaptors::QuickBooksOnline::LedgerClass::Operations::Update do
+  include AdaptorHelpers
+
+  let(:ledger_class) { LedgerSync::LedgerClass.new(ledger_id: '123', name: 'Test', active: true, sub_class: false)}
+
+  it do
```
ryanwjackson: We don't need to do this in this PR, but we could DRY up a lot of these specs using rspec shared examples, given that not much changes here.

ryanwjackson:
@mattgmarcus Want to review?

lib/ledger_sync/adaptors/quickbooks_online/department/ledger_serializer.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+module LedgerSync
+  module Adaptors
+    module QuickBooksOnline
+      module Department
+        class LedgerSerializer < QuickBooksOnline::LedgerSerializer
+          id
```
mattgmarcus: I'm unfamiliar with the DSL. Out of curiosity, if we can do this with `id`, can we do it with attributes like `name` and `active` too?

ryanwjackson: @mattgmarcus this is just a class function on the `QuickBooksOnline::LedgerSerializer` class given that most QBO serializers use IDs the same way.  We may change this given that we need to split up the serializers and deserializers anyway.

PR #114, opened 2019-12-22
Add classes for retrieving NetSuite record metadata

### LINKED ISSUE #113: NetSuite REST API record schema
The attributes of any given record needs to be retrieved from the API.  There are also custom fields that need to be accounted for.

Add a way to use the adaptor to pull the record schemas.

ryanwjackson:
Fix #113 

I wanted an easier way to retrieve and eventually work with NetSuite record (resource) attributes.  These classes essentially give us the documentation (request methods, attributes, etc.) as they vary user-to-user based on permissions.

The next step here will be to explore how we want to handle customer fields.

@mattgmarcus @jozefvaclavik please review.


lib/ledger_sync/adaptors/netsuite/record/metadata.rb
```diff
@@ -0,0 +1,70 @@
...
+          def initialize(adaptor:, record:)
+            @adaptor = adaptor
+            @record = record
+          end
+
+          def http_methods
+            @http_methods ||= begin
+              ret = []
+
+              http_methods_response.body['paths'].each do |path, path_data|
```
jozefvaclavik: I get that we're doing 2 dimensional array here, but isnt doing `.map` and `.map` and then `.flatten` still nicer?

PR #115, opened 2019-12-23
Add Invoice, Payment, LineItems and Txn resources and serializers

jozefvaclavik:
Instead of doing magic with Invoice line item type, I decided to represent different type with different resource. So we can do `InvoiceSalesLineItem`, `InvoiceDiscountLineItem`, etc.  The advantage is that each resource has different attributes, so we don't have to worry about trying go fit it all under one resource and serializer.

----

No more Txn resource. Instead `ledger_transactions: []` takes `Invoice` (and other) objects directly.


lib/ledger_sync/adaptors/quickbooks_online/item/ledger_serializer.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module LedgerSync
+  module Adaptors
+    module QuickBooksOnline
+      module Item
+        class LedgerSerializer < QuickBooksOnline::LedgerSerializer
```
ryanwjackson: Is there not more to this object?

lib/ledger_sync/resources/txn.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module LedgerSync
+  class Txn < LedgerSync::Resource
```
ryanwjackson: \+ @mattgmarcus 

Are we ok with this name?  I know `transaction` is a reserved word (we stupidly named our base ledger entry `transaction` at Paid, and it caused nightmares).  But like with `LedgerClass`, do we want to make this `LedgerTransaction`?

lib/ledger_sync/adaptors/quickbooks_online/txn/ledger_serializer.rb
```diff
@@ -0,0 +1,23 @@
...
+module LedgerSync
+  module Adaptors
+    module QuickBooksOnline
+      module Txn
+        class LedgerSerializer < QuickBooksOnline::LedgerSerializer
+          attribute ledger_attribute: 'TxnId',
+                    resource_attribute: 'entity.ledger_id'
+
+
+          attribute ledger_attribute: 'TxnType' do |resource|
```
ryanwjackson: Would it be better to put this in a `LedgerSerializerType`?

lib/ledger_sync/adaptors/quickbooks_online/item/ledger_serializer.rb
jozefvaclavik: We dont sync it. All we need is to reference relationship by ID.

lib/ledger_sync/resources/txn.rb
jozefvaclavik: I'm fine with `LedgerTransaction`

mattgmarcus: Sounds good to me

spec/adaptors/quickbooks_online/operations/invoice/create_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+support :input_helpers,
+        :operation_shared_examples,
+        :quickbooks_online_helpers
+
+RSpec.describe LedgerSync::Adaptors::QuickBooksOnline::Invoice::Operations::Create do
+  include InputHelpers
+  include QuickBooksOnlineHelpers
+
+  let(:customer) do
+    LedgerSync::Customer.new(customer_resource(ledger_id: '123'))
```
ryanwjackson: Please use `FactoryBot` so that we ensure variance does not matter.  And then we can get rid of the resource hashes.

lib/ledger_sync/adaptors/quickbooks_online/ledger_serializer_type/transaction_reference_type.rb
```diff
@@ -0,0 +1,58 @@
...
+              ret = resource_class.new(
+                ledger_id: item['TxnId']
+              )
+            end
+          end
+
+          def convert_from_local(value:)
+            return if value.nil?
+            raise "List expected.  Given: #{value.class.name}" unless value.is_a?(Array)
+            # raise "Resources expected.  Given: #{value.map{|i| i.class.name}.join(', ')}" unless value.map{|i| i.class}.all?(LedgerSync::Resource)
```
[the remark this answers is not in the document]
jozefvaclavik: So this was the interesting part. I basically took this from another type you wrote. And you compare that object is a `LedgerSync::Resource`, but in this case it was failing, because object is `LedgerSync::Invoice`. So wanted to run this by you.

spec/adaptors/quickbooks_online/operations/invoice/create_spec.rb
jozefvaclavik: Tbh for anything that adds `ledger_id` we dont need any factories. We need that object only as a `Ref`

PR #116, opened 2019-12-23
Add support for FactoryBot in tests

### LINKED ISSUE #32: Use factories for tests
Currently we have to define our own objects in `inputs_helper.rb` which is cumbersome and inflexible.

- Find a factory library that works without persistence/ORM.
- Absent a better alternative, use FactoryBot: https://thoughtbot.com/blog/tips-for-using-factory-girl-without-an-orm

- Writing our own factory library

ryanwjackson:
Fix #32


ryanwjackson:
@jozefvaclavik This doesn't rewind the sequences, because they aren't getting registered.  Do you see anything non-standard or something that jumps out as to why?

`FactoryBot` defines two `sequence` methods, one that registers the sequence globally and one that doesn't.  This is obviously using the latter because `FactoryBot.sequences` is empty.

ryanwjackson:
I monkey-patched the sequence issue and documented.  I didn't think it worth diving into.

I did not create detailed factories yet, but @jozefvaclavik let's start using factories for our examples so we can phase out `InputHelpers`.

Please merge on approval.

spec/factories/bills.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :bill, class: LedgerSync::Bill do
+    references_one :account
+
+    sequence(:memo) { |n| "Memo #{n}" }
```
mattgmarcus: Have you ever used the `Faker` gem? It could be nice for parts like the memo. But this is also fine for now

spec/factory_bot_spec.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
```
mattgmarcus: IMO you don't really need a test file for factorybot, since it'll be pretty extensively tested (albeit implicitly) through the rest of our tests.

spec/support/vcr.rb
```diff
@@ -0,0 +1,17 @@
...
+  config.hook_into :webmock
+  config.configure_rspec_metadata!
+
+  # These need to match the defaults in support/netsuite_soap_helpers.rb for the
+  # stubs to work.
+  config.filter_sensitive_data('netsuite_account_id') { ENV['NETSUITE_ACCOUNT_ID'] } if ENV.key?('NETSUITE_ACCOUNT_ID')
+  config.filter_sensitive_data('NETSUITE_CONSUMER_KEY') { ENV['NETSUITE_CONSUMER_KEY'] } if ENV.key?('NETSUITE_CONSUMER_KEY')
+  config.filter_sensitive_data('NETSUITE_CONSUMER_SECRET') { ENV['NETSUITE_CONSUMER_SECRET'] } if ENV.key?('NETSUITE_CONSUMER_SECRET')
+  config.filter_sensitive_data('NETSUITE_TOKEN_ID') { ENV['NETSUITE_TOKEN_ID'] } if ENV.key?('NETSUITE_TOKEN_ID')
+  config.filter_sensitive_data('NETSUITE_TOKEN_SECRET') { ENV['NETSUITE_TOKEN_SECRET'] } if ENV.key?('NETSUITE_TOKEN_SECRET')
```
mattgmarcus: Just double checking the capitalization and param names are all correct here, since line 12 is inconsistent

spec/factory_bot_spec.rb
ryanwjackson: This is to ensure our monkey patching of the rewinding continues to work, otherwise we would get inconsistent results which may be hard to debug.

spec/factories/bills.rb
ryanwjackson: I have used it.  We could start to do it.  I left it like this though in case you needed to line up errors with objects in the ledger itself or something.  Though uniqueness becomes an issue since tests will re-use the same names.  Happy either way.

spec/support/vcr.rb
ryanwjackson: I prefer CAPS for subs, but because the `account_id` is passed through the `URI` lib, it is downcased automatically.  So tests fail naturally or don't match the cassettes given the case.  VCR recommends doing something like `<NETSUITE_ACCOUNT_ID>` (with the brackets), but that causes failures given the variables are used in the headers and URL.

Issue #131, opened 2020-01-16
NetSuite Chart of Accounts


ryanwjackson:
**Is your feature request related to a problem? Please describe.**
Add support for managing accounts, departments, and classes in NetSuite to get a full chart of accounts.

PR #132, opened 2020-01-16
Add account operations for NetSuite

ryanwjackson:
Rel: #131


lib/ledger_sync/adaptors/mixins/offset_and_limit_pagination_searcher_mixin.rb
```diff
@@ -0,0 +1,37 @@
...
+          return nil if offset <= 1
+
+          paginate(limit: limit, offset: offset - limit)
+        end
+
+        private
+
+        def default_offset
+          0
+        end
```
mattgmarcus: Does NS start at 0?

It may be better to just raise in here (and `default_limit`) and force the class that includes the mixin to define it. I'm not sure if one value will truly be a default. So it'd be more like an interface

lib/ledger_sync/adaptors/netsuite/account/ledger_deserializer.rb
```diff
@@ -0,0 +1,17 @@
...
+module LedgerSync
+  module Adaptors
+    module NetSuite
+      module Account
+        class LedgerDeserializer < NetSuite::LedgerSerializer
+          attribute ledger_attribute: :id,
+                    resource_attribute: :ledger_id
+
+          attribute ledger_attribute: :acctname,
+                    resource_attribute: :name
```
mattgmarcus: While we're here, are there other attributes we can put in here? Similar to the attributes we have for account

Or do they get filled in automagically (I don't recall how that works)

lib/ledger_sync/adaptors/netsuite/account/ledger_serializer.rb
```diff
@@ -0,0 +1,38 @@
...
+
+          # # attribute ledger_attribute: :firstName,
+          # #           resource_attribute: :first_name
+
+          # # attribute ledger_attribute: :lastName,
+          # #           resource_attribute: :last_name
+
+          # attribute ledger_attribute: :subsidiary,
+          #           resource_attribute: :subsidiary,
+          #           type: LedgerSerializerType::ReferenceType
```
mattgmarcus: Do we not need these?

ryanwjackson:
@mattgmarcus This is just missing `currency`.  If you want to approve and merge, I can add it after we make the necessary changes to support a currency resource.

Issue #126, opened 2020-01-14
Quickbooks Online `revoke_token!` method throwing exception


SeanBolt:
**Describe the bug**
Quickbooks Online adaptor `revoke_token!` method throwing exception `An empty string is not a valid JSON string.`

**To Reproduce**
- Initialize quickbooks online adaptor
- Call `revoke_token!` method

**Expected behavior**
Quickbooks Online token to be revoked

**Additional context**
I dug into the oauth2 library and it seems the cause of this is a `401 Unauthorized` error

```
[...]

PR #133, opened 2020-01-17
Update revoke_token method to use basic auth

SeanBolt:
Fix: #126 

Use Basic Auth when revoking qbo token.


ryanwjackson:
@SeanBolt You will need to update this.  I was going to do it for you, but I can't figure out how to checkout and modify your branch in place.  Lmk if you need help.

lib/ledger_sync/adaptors/quickbooks_online/adaptor.rb
```diff
@@ -101,14 +101,15 @@ def refresh!
...
             body: {
               token: access_token
-            }.to_json,
-            headers: OAUTH_HEADERS.dup
-          ).status == 200
+            },
+            headers: headers,
+            method: :post,
+            url: REVOKE_TOKEN_URI
+          ).perform.status == 200
```
ryanwjackson: I think it probably best to return the request here for debugging purposes actually.  But I'll go ahead and approve and we can do that downstream if we want.

SeanBolt:
@ryanwjackson thanks for taking a look at this, I fixed the merge conflict so this looks good to merge in

PR #137, opened 2020-01-22
Currency Resource

### LINKED ISSUE #135: Add currency as resource
Currency is a resource in the NetSuite world.  It therefore can be CRUD managed, so to be "more correct," we need to support it as a `references_on` relationship where currency is used.

ryanwjackson:
# **THIS IS A BREAKING CHANGE**

Fix #135 

Currency is a full resource in NetSuite and a reference object in QBO.  You will need to change `resource.currency = 'USD'` to `resource.currency = LedgerSync::Currency.new(symbol: 'USD')`.

This gives us the ability to also pass `name`, CRUD operations in NetSuite, and more flexibility in associations.


lib/ledger_sync/adaptors/quickbooks_online/account/ledger_serializer.rb
```diff
@@ -5,15 +5,28 @@ module Adaptors
...
+                    resource_attribute: :name
+          attribute ledger_attribute: 'AccountType',
+                    resource_attribute: :account_type,
+                    type: LedgerSerializerType::AccountType
+          attribute ledger_attribute: 'AccountSubType',
+                    resource_attribute: :account_sub_type,
+                    type: LedgerSerializerType::AccountSubType
+          attribute ledger_attribute: 'AcctNum',
+                    resource_attribute: :number
+          attribute ledger_attribute: :CurrencyRef,
```
jozefvaclavik: If currency is a resource, shouldnt this be a relationship with `ledger_id: ‘USD’`?

ryanwjackson: It is handled in the type.

jozefvaclavik: I know, but its a relationship. Seems like we're introducing yet another another pattern how we set value. I think if we take away the dogma that ID is a numeric value, we can easily justify `LedgerSync::Currency.new(ledger_id: 'USD', name: 'United States Dollar')`.

And even if some ledger will use numeric ID for a currency instead of `USD`, it will be already done. Coz we will just do `LedgerSync::Currency.new(ledger_id: 123, name: 'United States Dollar')`

jozefvaclavik: Ok, to be fair, saying for QBO that currency has `ledger_id: 'USD'` when almost all other resources has numeric value is introducing complexity. Tho classes has weird IDs.

ryanwjackson: I'm not sure I agree with the ID stuff, though to be fair, I don't think we can assume it's numeric.  I think we most always convert to a string anyway.  

That said, I do agree that using the type was adding duplicate ways of achieving something.  I just switched it to using a currency serializer (similar to how we do with other `references_many` relationships).

That work?

ryanwjackson:
@mattgmarcus I did a find & replace all, so there are no longer `attribute :currency` in any resource and it's all `references_one :currency`

PR #138, opened 2020-01-22
Add options to disable de/serialize for any attribute

### LINKED ISSUE #121: Add option to LedgerSerializer for disabling de/serialize
Some APIs (e.g. NetSuite) do not use the same request and response bodies.

Add an option when creating an attribute to disable it for `serialize` or `deserialize` (a.k.a. so it only is used one direction.

- We can either use separate serializers and deserializers (perhaps the best longterm solution).

ryanwjackson:
Fix #121 

You can now disable serializing or deserializing specific attributes.  This will prevent us from rewriting serializers yet.

### DIFF
--- a/lib/ledger_sync/adaptors/ledger_serializer_attribute.rb
+++ b/lib/ledger_sync/adaptors/ledger_serializer_attribute.rb
@@ -7,27 +7,45 @@ module LedgerSync
   module Adaptors
     class LedgerSerializerAttribute
       attr_reader :block,
+                  :deserialize,
                   :id,
                   :ledger_attribute,
                   :resource_attribute,
                   :resource_class,
+                  :serialize,
                   :serializer,
                   :type
 
-      def initialize(block: nil, id: false, ledger_attribute:, resource_attribute: nil, resource_class: nil, serializer: nil, type: LedgerSerializerType::ValueType)
+      def initialize(
+        block: nil,
+        deserialize: true,
+        id: false,
+        ledger_attribute:,
+        resource_attribute: nil,
+        resource_class: nil,
+        serialize: true,
+        serializer: nil,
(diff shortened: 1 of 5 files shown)


lib/ledger_sync/adaptors/ledger_serializer_attribute.rb
```diff
@@ -30,6 +44,10 @@ def block_value_for(resource:)
         block.call(resource)
       end
 
+      def deserialize?
```
mattgmarcus: I thought you got these for free. Only in rails land?

lib/ledger_sync/adaptors/ledger_serializer_attribute_set.rb
```diff
@@ -4,9 +4,11 @@ module LedgerSync
   module Adaptors
     class LedgerSerializerAttributeSet
       attr_reader :attributes,
+                  :deserialize_attributes,
```
mattgmarcus: I might call these `deserialized_attributes` and `serialized_attributes` (with a `d`), since it sounds like a method otherwise. Just my take tho

lib/ledger_sync/adaptors/ledger_serializer_attribute.rb
ryanwjackson: Yeah. :(

lib/ledger_sync/adaptors/ledger_serializer_attribute_set.rb
ryanwjackson: What about `deserializable_attributes`?  I also feel like `deserialized_attributes` implies it was already done. haha

ryanwjackson: If cool with it, approve and merge. :)

PR #139, opened 2020-01-23
Add custom attributes for resources

### LINKED ISSUE #117: Custom resource attributes
NetSuite (and other ledgers) support customer fields/attributes on resources.  Right now, you must specify each attribute per resource (and in operation validations).

**Thoughts/Ideas**

1. `ResourceAttributeSet`
Right now we store all attributes in a `ResourceAttributeSet` object.  We could modify resources to have an `.attributes` instance method instead of attributes being directly accessible.  Then we can have that set accept any key/value pair.

2. Add a `custom_attributes: {}` keyword to resource initializer

3. Add a way to specify custom attributes when calling `LedgerSync.register_resource` (or somewhere when instantiating LedgerSync)

This would require devs to know and specify their custom attributes.  NetSuite does provide a dynamic schema/metadata through their API.  I don't know that this holds for other ledgers that offer this level of customization.  In theory, they may just return custom attributes as part of the record's response body and we may not have a formal way to detect.  So perhaps being intentional and having devs define these attributes makes sense.

4. Add raw body to resource

Ideally, resources will remain adaptor independent.  But there is data loss (even now) for all the attributes we aren't defining.  You can access this data through the `Response` object body, but it's pretty inconvenient.  Perhaps we can add an `adaptor_body` attribute to resources or something?

**Appendix**
For all of these we would have to determine how to handle operation validations, which right now require that you define a validation for every attribute.

ryanwjackson:
Fix #117


ryanwjackson:
Closing this, as we are going with a different solution discussed in #117

PR #141, opened 2020-01-23
Add Department and Class to Bill, Deposit, Expense and Journal Entry

jozefvaclavik:
Adds references to `Class` and `Department` to `Bill`, `BillLineItem`, `DepositLineItem`, `Expense`, `ExpenseLineItem` and `JournalEntryLineItem`.


lib/ledger_sync/adaptors/quickbooks_online/bill/ledger_serializer.rb
```diff
@@ -30,6 +30,9 @@ class LedgerSerializer < QuickBooksOnline::LedgerSerializer
           attribute ledger_attribute: 'APAccountRef.value',
                     resource_attribute: 'account.ledger_id'
 
+          attribute ledger_attribute: 'DepartmentRef.value',
+                    resource_attribute: 'department.ledger_id'
```
ryanwjackson: I know you were for only having `attribute`, but we do have the following:

```ruby
references_one :department, serializer: Department::LedgerSerializer
```

If the structure of the sub-hash is the same, it can handle parsing everything (including not just ID).

This goes back to your old comment of not necessarily having the dot-strings to handle these cases and using nested serializers.

Issue #140, opened 2020-01-23
Multiple operations not syncing new resources


ryanwjackson:
**Describe the bug**
```ruby
adaptor = LedgerSync::Adaptors::QuickBooksOnline::Adaptor.new(access_token: '', refresh_token: '', realm_id: '', client_id: '', client_secret: '', test: true)
resource1 = LedgerSync::Vendor.new(display_name: 'Ryan 1')
operation1 = LedgerSync::Adaptors::QuickBooksOnline::Vendor::Operations::Create.new(adaptor: adaptor, resource: resource1)
resource2 = LedgerSync::Vendor.new(display_name: 'Peter 1')
operation2 = LedgerSync::Adaptors::QuickBooksOnline::Vendor::Operations::Create.new(adaptor: adaptor, resource: resource2)
operation1.perform
operation2.perform
```

```
operation1:
url: https://sandbox-quickbooks.api.intuit.com/v3/company/123146386787439/vendor
body: {"Id":null,"DisplayName":"Ryan 1","GivenName":null,"FamilyName":null,"PrimaryEmailAddr":{"Address":null}}
response: {"Vendor":{"Balance":0,"Vendor1099":false,"CurrencyRef":{"value":"USD","name":"United States Dollar"},"domain":"QBO","sparse":false,"Id":"85","SyncToken":"0","MetaData":{"CreateTime":"2020-01-23T08:19:31-08:00","LastUpdatedTime":"2020-01-23T08:19:31-08:00"},"DisplayName":"Ryan 1","PrintOnCheckName":"Ryan 1","Active":true},"time":"2020-01-23T08:19:31.825-08:00"}
operation2:
url: https://sandbox-quickbooks.api.intuit.com/v3/company/123146386787439/vendor
body: {"Id":null,"DisplayName":"Ryan 1","GivenName":null,"FamilyName":null,"PrimaryEmailAddr":{"Address":null}}
```

PR #144, opened 2020-01-24
Update QBO Stubs

### LINKED ISSUE #143: Make QBO stubbing more scalable
When we add new resources/attributes, if they are tied into QBO, we have to edit each individual body and response string to match accordingly.  Because they are strings, they must match in order as well.  It adds significant time to update tests

Use hashes and `to_json` for standardizing QBO API JSON objects.

- vcr recording, but this requires connecting to the live QBO instance.  While this is arguably "more accurate," it can be challenging to create the right scenario to record in a single unit/feature test.

ryanwjackson:
Fix #143 

When I was adding currency, updating all of the stubs in QBO that reference them was such a pain.  I started to updated how we stub in QBO, making it a bit more flexible/extensible.

One thing of note is I added a forced expectation on the number of times a stub is to be called.  This would have saved us from #140 given that we would have only expected that one stub to be called once.  At a minimum, I think we should keep this.

For the rest, I wanted to get some feedback before I continue to convert the rest.  I would say we can merge this now and migrate slowly, so that we don't have a lot of merge conflicts.

@jozefvaclavik @mattgmarcus thoughts?


jozefvaclavik:
I'm hesitant.

On one side youre saying: we have strings, if we wanna update, we need to update whole string. - which is not always true, you just need to update part of a string you updated.

On the other you create hash, and if we need to update it, we need to update part of the hash hash. if youre working with new resource, youre gonna get back response body as a string, so you will need to convert it to hash and format it. 

Is hash more readable? Totally. Will we need to pull updated response body and match it against our hash once QBO structure changes? Totally. Thinking of having class/department in QBO response once you enable them in the QBO.

Tbh its not like we are stubbing one response in multiple resources.
We have 2 different request body:
1. create resource - consists from our hash
2. update resource - merged QBO and our hash

And we have 1 response body thats wrapped in two different root keys:
1. object response - returned for create, find and update operation - update increments `SyncToken`, but we don't validate that so we can re-use one response - this is usually: `{"Customer":BODY}`
2. search response - wrapped in array - `{"QueryResponse":{"Customer":[BODY]}}`

So to save time updating a stubs, we could easily just reuse one response body in all the responses. *(caugh) I did that few times. (caugh)*

ryanwjackson:
I understand what you are saying.  I tend to like to do things explicitly, but my issues/thoughts are the following:

0. Practically, it took me ~1 hour to update tests just to add currency support as it became a resource and added new attributes.  I had to find and replace all instances, which I did through the editor.  Then I relied on tests failing.  But tbh, I'm not sure if I updated it everywhere, and sometime downstream, that may matter.

1. It requires us to remember to require updating every place or adding every new place for coverage.  Some routes will not consider certain fields, but perhaps they would matter and we just forgot to include or update them.

2. In theory (big statement here considering what we've seen), bodies of requests and responses should be the same.  We can DRY up our code.

3. Even though we are working with requests/responses, I more consider these unit tests.  I think it fair to assume that we can define the body once, as we are just using them (99% of the time) to test the operation.  I'm OK letting the QA spec set be our "integration" tests, as they will hit live APIs and if something fails, then we will need to update bodies in unit tests anyway.

4. We do need to add more tests, especially around the expected output of operations (e.g. did it actually update the name of the returned resource?).  We also can/should do this in our QA tests.

5. As with the #142, we are introducing instances where we need to modify the body of a response explicitly.  We can either copy a string and re-write it (introducing more duplication) or do `JSON.parse(body).merge(...).to_json` (which seems like overkill) or just have hashes to begin with.

6. I am not 100% sold on how I wrote this yet, especially given the whole `FullUpdate` thing.  I only did one resource/set of operations, so I expect that to change.

7. Again, I prefer being explicit as well, but I more firmly believe we need to enforce certain tests.   `it_behaves_like 'an operation'` ensures that we don't forget to write critical tests.  The dynamic definition of stub methods ensures they are consistent.  The `times` stub checking ensures we don't have issues like #142.  Not all tests need to fall in those buckets, and our specs should be more diverse than they are now (or we could just automagically test everything anyway).

Sorry, that's a lot of thoughts, but it all stems from two main areas:

1. Development speed
2. Test coverage/completeness

ryanwjackson:
One last thing (which is where the frustration stemmed from):

If you change a string, the stubbing fails unless the string matches (not the parsed string).  For example:

```
diff --git a/spec/support/quickbooks_online_helpers.rb b/spec/support/quickbooks_online_helpers.rb
index 19a8283..32b673e 100644
--- a/spec/support/quickbooks_online_helpers.rb
+++ b/spec/support/quickbooks_online_helpers.rb
@@ -732,7 +732,7 @@ module QuickBooksOnlineHelpers
   def stub_update_transfer
     stub_request(:post, 'https://sandbox-quickbooks.api.intuit.com/v3/company/realm_id/transfer')
       .with(
-        body: '{"Id":"123","Amount":123.45,"PrivateNote":"Memo","FromAccountRef":{"value":"123","name":"Sample Account"},"ToAccountRef":{"value":"123","name":"Sample Account"},"TxnDate":"2019-09-01","CurrencyRef":{"value":"USD","name":"United States Dollar"},"domain":"QBO","sparse":false,"SyncToken":"0","MetaData":{"CreateTime":"2019-10-03T21:46:28-07:00","LastUpdatedTime":"2019-10-03T21:46:28-07:00"}}',
+        body: '{"Id":"123","Amount":123.45,"PrivateNote":"Memo","FromAccountRef":{,"name":"Sample Account","value":"123"},"ToAccountRef":{"value":"123","name":"Sample Account"},"TxnDate":"2019-09-01","CurrencyRef":{"value":"USD","name":"United States Dollar"},"domain":"QBO","sparse":false,"SyncToken":"0","MetaData":{"CreateTime":"2019-10-03T21:46:28-07:00","LastUpdatedTime":"2019-10-03T21:46:28-07:00"}}',
         headers: {
           'Accept' => 'application/json',
           'Accept-Encoding' => 'gzip;q=1.0,deflate;q=0.6,identity;q=0.3',
@@ -743,7 +743,7 @@ module QuickBooksOnlineHelpers
       )
       .to_return(
         status: 200,
-        body: '{"Transfer":{"FromAccountRef":{"name":"Sample Account","value":"123"},"ToAccountRef":{"value":"123","name":"Sample Account"},"Amount":123.45,"domain":"QBO","sparse":false,"Id":"123","SyncToken":"1","MetaData":{"CreateTime":"2019-10-03T21:46:28-07:00","LastUpdatedTime":"2019-10-03T21:46:28-07:00"},"TxnDate":"2019-09-01","CurrencyRef":{"value":"USD","name":"United States Dollar"},"PrivateNote":"Memo"},"time":"2019-10-03T21:46:27.963-07:00"}',
+        body: '{"Transfer":{"FromAccountRef":{"value":"123","name":"Sample Account"},"ToAccountRef":{"value":"123","name":"Sample Account"},"Amount":123.45,"domain":"QBO","sparse":false,"Id":"123","SyncToken":"1","MetaData":{"CreateTime":"2019-10-03T21:46:28-07:00","LastUpdatedTime":"2019-10-03T21:46:28-07:00"},"TxnDate":"2019-09-01","CurrencyRef":{"value":"USD","name":"United States Dollar"},"PrivateNote":"Memo"},"time":"2019-10-03T21:46:27.963-07:00"}',
         headers: {}
       )
   end
```

i just changed `{"value":"123","name":"Sample Account"}` to `{"name":"Sample Account","value":"123"}` in the string.

The stub no longer matches, even though `JSON.parse(s1) == JSON.parse(s2)`.  If you use a hash of the body and call `body_hash.to_json`, then it matches.

I added `spec/stubbing_spec.rb` to demo  (note: the last expects an error): https://github.com/LedgerSync/ledger_sync/pull/144/files#diff-e4e886a31b2248aac7162129627b1c39

jozefvaclavik:
After having a discussion with @ryanwjackson we agreed that each of us takes the specs to be something else.

As our integration tests are not mocking exact response from QBO (they never really did as I always updated related ids to match `123`, etc), we can treat them as a fancy serializer tests.

1. given resource input, serialized output matches request we're expecting to stub.
2. given stub response, deserialized resource has values were expecting.
3. hidden benefit is that we test that adaptor actually tries to make a request

mattgmarcus:
I'm fine with wherever you two end up, so I'll leave this to you guys

ryanwjackson:
@mattgmarcus @jozefvaclavik @SeanBolt Fixed merge conflicts.  Please merge on approval.

PR #147, opened 2020-01-28
Custom Attributes

### LINKED ISSUE #117: Custom resource attributes

ryanwjackson:
Fix #117 

@mattgmarcus @jozefvaclavik this is ready for review.  I added examples to the README, but please let me know if that's not clear.


README.md
```diff
@@ -96,18 +94,67 @@ Resources have defined attributes.  Attributes are explicitly defined.  An error
...
 
+#### Custom Attributes
+
+Some ledgers (e.g. NetSuite) allow for custom attributes (or "fields"), which can vary by account and user.  To allow for custom attributes, you can create new resources, ledger serializers (see below), and validation contracts (see below).  Assuming your ledger supports the string attribute `foo` for customers, you could do the following:
+
+```ruby
+class CustomCustomer < LedgerSync::Customer
+  attribute :foo, type: LedgerSync::Type::String
+end
+```
```
mattgmarcus: Can you put an example of setting this like this for documentation
```
klass = Class.new(LedgerSync::Customer)
klass.attribute :foo, type: LedgerSync::Type::String

```

README.md
```diff
@@ -96,18 +94,67 @@ Resources have defined attributes.  Attributes are explicitly defined.  An error
...
+Operations depend on `LedgerSync::Adaptor::LedgerSerializer`s to serialize and deserialize objects.  Most resources have a default serializer per adaptor which may acts as both a serializer and deserializer.  You can provide custom serializers, which is necessary when working with custom attributes.  For example, given the following:
+- `custom_resource` that is a `LedgerSync::Customer` (see above)
+- the attribute `foo` is used in both the request and response bodies
+- using the `NetSuite` adaptor
+
+you could implement custom serializers using the following code:
+
+```ruby
+# Extend the existing one or use
+# LedgerSync::Adaptors::NetSuite::LedgerSerializer or, more generically,
```
mattgmarcus: Should `LedgerSync::Adaptors::NetSuite::LedgerSerializer` be `LedgerSync::Adaptors::NetSuite::Customer::LedgerSerializer` ?

lib/ledger_sync/adaptors/ledger_serializer.rb
```diff
@@ -12,9 +12,8 @@ class LedgerSerializer
 
       attr_reader :resource
 
-      def initialize(ensure_inferred_resource_class: true, resource:)
+      def initialize(resource:)
```
mattgmarcus: We had talked about maybe altering this in such a way that this interface doesn't break going forward. Up to you

lib/ledger_sync/adaptors/operation.rb
```diff
@@ -46,16 +48,34 @@ def self.included(base)
...
-          raise 'Missing resource' if resource.nil?
-
-          raise "#{resource.class.name} is not a valid resource type.  Expected #{self.class.resource_klass.name}" unless resource.is_a?(self.class.resource_klass)
+        def initialize(
+          adaptor:,
+          ledger_deserializer_class: nil,
+          ledger_serializer_class: nil,
+          resource:,
+          validation_contract: nil
+        )
```
mattgmarcus: Same point here about interface

lib/ledger_sync/adaptors/operation.rb
```diff
@@ -46,16 +48,34 @@ def self.included(base)
...
+          self.class.raise_if_unexpected_class(expected: self.class.resource_klass, given: resource.class)
 
           @adaptor = adaptor
+
+          unless ledger_deserializer_class.nil?
```
mattgmarcus: `if x.present?` seems easier to me than reading `unless x.nil?` IMO

lib/ledger_sync/error.rb
```diff
@@ -1,10 +1,43 @@
+# frozen_string_literal: true
+
 module LedgerSync
   class Error < StandardError
+    module HelpersMixin
+      module ClassMethods
+        def raise_if_unexpected_class(expected:, given:)
+          expected = [expected] unless expected.is_a?(Array)
```
mattgmarcus: `expected = Array.wrap(expected)` I think basically does this, if you're interested in using that. Doesn't matter though

README.md
ryanwjackson: Not necessarily.  The inheritance looks like this:

` LedgerSync::Adaptors::NetSuite::Customer::LedgerSerializer <  LedgerSync::Adaptors::NetSuite::LedgerSerializer <  LedgerSync::Adaptors::LedgerSerializer`

It really just depends on what you're looking to accomplish.

I updated the readme on this to hopefully be a bit more clear.

lib/ledger_sync/adaptors/operation.rb
ryanwjackson: If it's `nil` it defaults it (hence the function call two lines below).  If it's present, then it does a type check

ryanwjackson:
@mattgmarcus @SeanBolt Please merge on approval.

PR #148, opened 2020-02-05
Add resource around Preferences

### LINKED ISSUE #107: Add resource to search preferences
https://developer.intuit.com/app/developer/qbo/docs/api/accounting/all-entities/preferences#read-preferences

I'd like the ability to pull out the preferences object from QBO, given a specific company.
This object is read only, since it is just a reflection of that company's preferences. I'm not sure if we want to handle this differently as a result. Like maybe we just return a hash of the data. This would be QBO specific.

Specifically I'd like to get the attributes `Preferences.AccountingInfoPrefs.ClassTrackingPerLine` and `Preferences.AccountingInfoPrefs.ClassTrackingPerTxn`

I'm not sure if their API docs are wrong by the way. For that link I put up top, for example, here's the names I see for the attributes
image

But then in the Purchase section (https://developer.intuit.com/app/developer/qbo/docs/api/accounting/all-entities/purchase), I see this different attribute `Preferences.AccountingInfoPrefs.ClassTrackingPerLine `
image

jozefvaclavik:
fix #107

There are few things that should be cleared out first

1. as a resource, Preference doesnt have any attribute - as they are all QBO specific
2. you can access it from `result = operation.perform` and `result.response.body` you will get full preferences response
3. QBO states there is no notion of `Id` for prefernces object, but then they go ahead and respond with `Id=1`. But you cannot use that to search for preferences object (ie: `/v3/company/123/preferences/1`). sigh.


lib/ledger_sync/adaptors/quickbooks_online/operation/find.rb
```diff
@@ -12,11 +12,11 @@ class Find
...
-            return failure(
-              Errors::OperationError::LedgerIDRequired.new(
-                operation: self
-              )
-            ) if resource.ledger_id.nil?
+            # return failure(
+            #   Error::OperationError::LedgerIDRequired.new(
+            #     operation: self
+            #   )
+            # ) if resource.ledger_id.nil?
```
jozefvaclavik: This had a typo for error. Anyway, do we need this check in a first place? when all our ledger operations defines `required(:ledger_id).filled(:string)`

spec/support/quickbooks_online_helpers.rb
```diff
@@ -2,6 +2,128 @@
 
 module QuickBooksOnlineHelpers
   STUBS = {
+    preferences: {
```
jozefvaclavik: Needs to be pluralized.

spec/support/quickbooks_online_helpers.rb
```diff
@@ -55,6 +177,9 @@ module QuickBooksOnlineHelpers
 
   def api_url(ledger_id: nil, resource:)
     ret = "https://sandbox-quickbooks.api.intuit.com/v3/company/realm_id/#{resource}"
+    if resource.to_s == 'preferences'
```
jozefvaclavik: Just a temporary hack untill we move some other resources and decide how we will do better resource to url mapping, coz we have few that are custom.

ryanwjackson:
@jozefvaclavik After some discussion, we should keep this as a resource, as it turns out you can update preferences, but what we would like to do is make it our first ledger-specific resource: `LedgerSync::Adaptors::QuickBooksOnline::Preferences`.  With this, we can actually pull out the preferences as attributes and operate on them on update.

This will probably (unless we removed it), require that we update the resource class validation when instantiating an operation (it checks the resource is the expected type).

So can we do the following:

- move the resource to under the QBO adaptor
- add the attributes
- add an update op based on https://developer.intuit.com/app/developer/qbo/docs/api/accounting/most-commonly-used/preferences#full-update-preferences
- Add QA tests so we know it works beyond stubbing

lib/ledger_sync/adaptors/quickbooks_online/operation/find.rb
ryanwjackson: I suppose it's redundant.  If it is only used here, can we just nuke any mention/definition of it?

lib/ledger_sync/adaptors/quickbooks_online/adaptor.rb
```diff
@@ -13,7 +13,8 @@ class Adaptor < LedgerSync::Adaptors::Adaptor
         ROOT_SANDBOX_URI  = 'https://sandbox-quickbooks.api.intuit.com'
         API_RESOURCE_MAPPING_OVERRIDE = {
           LedgerSync::Expense => 'purchase',
-          LedgerSync::LedgerClass => 'class'
+          LedgerSync::LedgerClass => 'class',
+          LedgerSync::Preference => 'preferences'
```
ryanwjackson: What do you think about naming it `Preferences` given that it is a collection of preferences?

lib/ledger_sync/resources/preference.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module LedgerSync
+  class Preference < LedgerSync::Resource
```
ryanwjackson: See my previous comment on the PR, but we should move this under the QBO adaptor and just have it be adaptor-specific so we can define the actual attributes.

spec/support/quickbooks_online_helpers.rb
ryanwjackson: Solved/OK if we rename to `Preferences` plural?

lib/ledger_sync/adaptors/quickbooks_online/adaptor.rb
jozefvaclavik: Yeah, that makes more sense.

PR #149, opened 2020-02-05
Add BillPayment resource

### LINKED ISSUE #60: Add resource support for BillPayment
https://developer.intuit.com/app/developer/qbo/docs/api/accounting/all-entities/billpayment

We need to be able to create, update, read, and delete bill payments, including using Searchers for this

- Id
- VendorRef
- Line (it looks like these reference `Bill` amongst other entities we won't worry about yet)
- TotalAmt
- PayType
- CurrencyRef
- PrivateNote
- TxnDate
- ExchangeRate
- APAccountRef

jozefvaclavik:
fix #60

Tried to implement the new quickbooks helper. I used real response from QBO, but some fields we're not present in my response (like `ARAccountRef`)

~Still missing `PayType`~


.tool-versions
```diff
@@ -0,0 +1 @@
+ruby 2.6.2
```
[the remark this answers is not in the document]
jozefvaclavik: https://asdf-vm.com/#/

spec/adaptors/quickbooks_online/operations/bill_payment/find_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+require 'spec_helper'
+
+support :operation_shared_examples,
+        :quickbooks_online_helpers
+
+RSpec.describe LedgerSync::Adaptors::QuickBooksOnline::BillPayment::Operations::Find do
+  include QuickBooksOnlineHelpers
+
+  let(:vendor) do
+    LedgerSync::Vendor.new(ledger_id: '123')
```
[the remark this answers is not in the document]
jozefvaclavik: I get that if we wanna mock better QBO response that sends also `name` attribute with `value` in `VendorRef`. But that may not be same depending on a factory. In case we do some internal caching of existing factories and we will sequence name automaticaly, etc.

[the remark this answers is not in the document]
jozefvaclavik: Allrite; all good here, scratch that last comment 👌

.tool-versions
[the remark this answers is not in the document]
ryanwjackson: Can you remove .tool-versions?  You can add it to your global git ignore: https://help.github.com/en/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer

ryanwjackson:
@jozefvaclavik Can we add some QA tests for this?  Just in the `/qa` directory, so we have some basic coverage against the actual API?

jozefvaclavik:
```
~/Repositories/mt/ledger_sync jv/bill_payment *8 !2 ❯ bundle exec rspec qa/quickbooks_online/bill_payment_spec.rb -I ./qa --require qa_helper.rb                                                                 20s   2.6.2 17:31:03
...

Finished in 28.67 seconds (files took 3.03 seconds to load)
3 examples, 0 failures

~/Repositories/mt/ledger_sync jv/bill_payment *8 !2 ❯                                                                                                                                                       32s   2.6.2 17:34:44
```

lib/ledger_sync/adaptors/quickbooks_online/bill_payment/ledger_serializer.rb
```diff
@@ -0,0 +1,76 @@
...
+          attribute ledger_attribute: 'TxnDate',
+                    resource_attribute: :transaction_date,
+                    type: LedgerSerializerType::DateType
+
+          attribute ledger_attribute: 'PayType',
+                    resource_attribute: :payment_type,
+                    type: LedgerSerializerType::PaymentType
+
+          attribute ledger_attribute: 'CreditCardPayment' do |res|
+            if res.credit_card_account
```
ryanwjackson: I wonder if we should use types for this, given we have to deserialize as well?  I am not sure block passing should exist and may just be a relic.  That said, I also don't know about the attribute structure here.  Once we dig into NS bill resource, we will have a clearer picture of the differences and how to model.

lib/ledger_sync/resources/bill_payment.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+require_relative 'account'
+require_relative 'currency'
+require_relative 'department'
+require_relative 'vendor'
+require_relative 'bill_payment_line_item'
+
+module LedgerSync
+  class BillPayment < LedgerSync::Resource
```
ryanwjackson: I don't know why I'm just now thinking of this, but is there a reason we wouldn't just use `Payment`?  The attributes look the same some attributes.  Is `Payment` one direction and this is the other?

ryanwjackson: @mattgmarcus I'm curious if you have thoughts on this.

On one hand it reminds me of bi-direction POs in MT.

On the other hand, it also seems similar to our current counterparty issue of having one resource that could be a vendor, customer, or employee.  Perhaps specificity at the resource class level is best.

jozefvaclavik: I think the relationship is more like `Invoice` -> `Payment`, `Bill` -> `BillPayment`

lib/ledger_sync/adaptors/quickbooks_online/bill_payment/ledger_serializer.rb
jozefvaclavik: True.

For `CreditCardPayment` inside there is just `CCAccountRef`. So its a simple reference.

For `CheckPayment` there is `BankAccountRef` and optional `PrintStatus` string attribute.

I may have to write type for it after all.

PR #164, opened 2020-02-28
Update for new NS version

ryanwjackson:


spec/qa/netsuite/customer_spec.rb
```diff
@@ -1,6 +1,10 @@
 # frozen_string_literal: true
 
-RSpec.describe LedgerSync::Adaptors::NetSuite::Customer, adaptor: :netsuite do
+require 'spec_helper'
+
+require 'spec_helper'
```
SeanBolt: Remove duplicate `require`

ryanwjackson:
@SeanBolt Should all be working now.

I don't expect them to change often, but it is in beta.  It's also not the only change, for example, they started requiring `externalId` for some (if not all) objects.  I assume that applies to all APIs and not just REST.

Issue #154, opened 2020-02-20
Add NetSuite Department resource


ryanwjackson:
- [x] operations
  - [x] create
  - [x] delete
  - [x] find
  - [x] update
- [x] searcher
- [x] qa tests

https://www.netsuite.com/help/helpcenter/en_US/srbrowser/Browser2019_2/script/record/department.html

SeanBolt:
@ryanwjackson are we planning to add Department to the `Netsuite adaptor`, `Netsuite Soap adaptor` or both?

ryanwjackson:
Just the netsuite rest adaptor for now.

PR #166, opened 2020-03-02
Add NetSuite Department Resource

SeanBolt:
#154  - Add NetSuite Department Resource


lib/ledger_sync/adaptors/netsuite/department/ledger_serializer.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module LedgerSync
+  module Adaptors
+    module NetSuite
+      module Department
+        class LedgerSerializer < NetSuite::LedgerSerializer
```
ryanwjackson: Confusing, I know, but in this case you don't need the deserializer if they are the same.  It will infer the "opposite" of the serializer.  We still need to clean that up, but just FYI.

lib/ledger_sync/adaptors/netsuite/department/operations/create.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+# require_relative '../../operation/create'
```
ryanwjackson: Remove?

lib/ledger_sync/adaptors/netsuite/department/operations/create.rb
```diff
@@ -0,0 +1,27 @@
...
+
+module LedgerSync
+  module Adaptors
+    module NetSuite
+      module Department
+        module Operations
+          class Create < NetSuite::Operation::Create
+            class Contract < LedgerSync::Adaptors::Contract
+              params do
+                required(:external_id).value(:string)
```
ryanwjackson: I could have done this wrong, but I think this should be `required(:external_id).filled(:string)` (not `value`.

lib/ledger_sync/adaptors/netsuite/department/searcher.rb
```diff
@@ -0,0 +1,28 @@
...
+      module Department
+        class Searcher < Searcher
+          def resources
+            @resources ||= begin
+              @request = adaptor
+                .get(
+                  path: "/department?limit=#{limit}&offset=#{offset}"
+                )
+
+              request.body
```
ryanwjackson: We can probably DRY this up with `inferred_resource_class` later on, seeing as it looks like most searchers follow this pattern.  Maybe even above, though I imagine we will need to make searchers more advanced (e.g. handle different attributes) at some point.

spec/adaptors/netsuite/department/operations/create_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+
+support :input_helpers,
+        :operation_shared_examples,
+        :netsuite_helpers
+
+RSpec.describe LedgerSync::Adaptors::NetSuite::Department::Operations::Create do
+  include InputHelpers
+  include NetSuiteHelpers
+
+  let(:resource) { LedgerSync::Department.new(department_resource) }
```
ryanwjackson: I prefer using `FactoryBot.create(:department)`, as this is the old way we used to create resources.

spec/adaptors/netsuite/department/operations/delete_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+
+support :input_helpers,
+        :operation_shared_examples,
+        :netsuite_helpers
+
+RSpec.describe LedgerSync::Adaptors::NetSuite::Department::Operations::Delete do
+  include InputHelpers
+  include NetSuiteHelpers
+
+  let(:resource) { LedgerSync::Department.new(department_resource(ledger_id: 7)) }
```
ryanwjackson: Use `FactoryBot.create(:department, ledger_id: 7)` instead.

spec/adaptors/netsuite/department/operations/find_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+
+support :input_helpers,
+        :operation_shared_examples,
+        :netsuite_helpers
+
+RSpec.describe LedgerSync::Adaptors::NetSuite::Department::Operations::Find do
+  include InputHelpers
+  include NetSuiteHelpers
+
+  let(:resource) { LedgerSync::Department.new(department_resource(ledger_id: 7)) }
```
ryanwjackson: `FactoryBot.create(:department, ledger_id: 7)`

spec/adaptors/netsuite/department/operations/update_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+
+support :input_helpers,
+        :operation_shared_examples,
+        :netsuite_helpers
+
+RSpec.describe LedgerSync::Adaptors::NetSuite::Department::Operations::Update do
+  include InputHelpers
+  include NetSuiteHelpers
+
+  let(:resource) { LedgerSync::Department.new(department_resource(ledger_id: 7)) }
```
ryanwjackson: `FactoryBot.create(:department, ledger_id: 7)`

lib/ledger_sync/adaptors/netsuite/department/operations/create.rb
SeanBolt: @ryanwjackson I was following what we do for ns `account`, `currency`, `customer` etc, which all use `.value(:string)`

After looking at the [dry-schema docs](https://dry-rb.org/gems/dry-schema/1.4/basics/macros/#value) I think you are right in saying `.filled` is what we should have.

ryanwjackson: We can do a find all and replace probably.  We can do it outside this PR so it doesn't block you.

ryanwjackson: Added to #167

lib/ledger_sync/adaptors/netsuite/department/searcher.rb
ryanwjackson: Added to #168

PR #172, opened 2020-04-09
Netsuite Invoice

ryanwjackson:
# Depends On

- [ ] https://github.com/LedgerSync/ledger_sync/pull/178


lib/ledger_sync/adaptors/netsuite/adaptor.rb
```diff
@@ -88,6 +88,12 @@ def self.ledger_attributes_to_save
           %i[]
         end
 
+        def self.ledger_resource_type_overrides
+          {
+            LedgerSync::Item => 'noninventoryitem'
```
ryanwjackson, line 93: This may be restrictive, but I'm not sure.  `item` is not a valid NS resource type, which suggests there may be more than one "specific" kind.

ryanwjackson: @tatecarr do you know the complete list of actual item types?  Is that pretty much fixed?

tatecarr: Slacked about this quick, short answer is the item sub types are pretty static, don't recall them changing the last few years. I would hope this is the complete list, but anywhere there is an item ref in the rest API it says item is "one of" and provides a list:

```
item: one of: [ inventoryItem, nonInventoryItem, serviceItem, otherChargeItem, assemblyItem, kitItem, nsResource, discountItem, markupItem, subtotalItem, descriptionItem, paymentItem, salesTaxItem, taxGroup, shipItem, downloadItem, giftCertificateItem, subscriptionPlan ]
```

ryanwjackson:
Close stale PR.

PR #180, opened 2020-04-16
DRY up collections

ryanwjackson:


lib/ledger_sync.rb
```diff
@@ -44,6 +44,7 @@
 require 'ledger_sync/util/performer'
 require 'ledger_sync/util/validator'
 require 'ledger_sync/util/string_helpers'
+require 'ledger_sync/util/mixins/delegate_array_methods_mixin'
```
jozefvaclavik, line 47: Arent all these `util` mixins? Whats different about this one that it needs `mixin` in the name and also in suffix?

ryanwjackson: No, some are utility classes.  And I guess the naming is a bit duplicated, but the convention was to make `Mixin` in the name if that's what it is.  Just scoped it given that most utils aren't mixins.

PR #187, opened 2020-04-21
Add .ruby-version

ryanwjackson:


jozefvaclavik:
Docker and travis still runs on 2.5.1. Should we bump those as erll?

ryanwjackson:
Good call.  Doing those now.  And not going 2.7.  There's a lot of warnings for various gems.  Put us on the latest 2.6.

--
Ryan Jackson
Product Manager
Modern Treasury ( https://www.moderntreasury.com )

PR #192, opened 2020-04-22
Use SuiteQL for NetSuite searchers

SeanBolt:
This PR moves the NetSuite searcher logic to use the `/suiteql` endpoint. The motivation for this change came about from the discovery that NetSuite index endpoints return a list of record ids without any further record information.


SeanBolt:
@ryanwjackson what do you think of this pattern for SuiteQL queries?

ryanwjackson:
I think it's fine.  I'm more just concerned about converting the response body into the resources.  Thoughts on that?

PR #194, opened 2020-04-23
DRY up tests with new shared examples.

ryanwjackson:
Used recent PR to DRY up tests.


spec/support/netsuite/shared_examples.rb
```diff
@@ -23,10 +34,13 @@
         stub_find_for_record
         stub_create_for_record
       when :delete
+        resource.ledger_id = netsuite_records.send(record).hash['id']
```
SeanBolt, line 37: Do we need to do this here since it's being done in the above `before` block?

ryanwjackson: Fixed.

PR #199, opened 2020-05-06
Add Reference serializer

jozefvaclavik:
Similar as `Currency::LedgerSerializer`

`references_one` now uses `Reference::LedgerSerializer` that returns:
1. `{value: related.ledger_id, name: related.name}` for existing relationship
2. `nil` if relationship is `nil`


ryanwjackson:
@jozefvaclavik should we just close this and add downstream?

jozefvaclavik:
Yeah, I would rather start from scratch then resolving all these conflicts.

PR #201, opened 2020-05-06
Use new serializers in NetSuite

ryanwjackson:


lib/ledger_sync/adaptors/netsuite/customer/serializer.rb
```diff
@@ -0,0 +1,34 @@
...
+
+          attribute :phone,
+                    resource_attribute: :phone_number
+
+          # attribute :firstName,
+          #           resource_attribute: :first_name
+
+          # attribute :lastName,
+          #           resource_attribute: :last_name
+
```
SeanBolt, line 27: Are we leaving these commented out blocks in?

lib/ledger_sync/adaptors/netsuite/adaptor.rb
```diff
@@ -60,6 +60,14 @@ def get(**keywords)
           request(keywords.merge(method: :get))
         end
 
+        def ledger_resource_path(args = {})
+          resource = args.fetch(:resource, nil)
+
+          ret = resource.class.resource_type.to_s # This can be turned into a case statement if we need to override
```
SeanBolt, line 66: How will we deal with `ledger_resource_type_overrides`?

lib/ledger_sync/adaptors/netsuite/customer/serializer.rb
ryanwjackson: These were in the originals, so I left them.  Not sure why though.  I figured I'd leave them for now until we decide to add/remove.

lib/ledger_sync/adaptors/netsuite/adaptor.rb
ryanwjackson: For now, we are going to have to just override here as needed with a case. The issue is that the current overrides only considers the class.  But the NS item for example will switch depending on a value of the item.

I'd say lave and then we perhaps combine concepts as we figure out exactly how to handle the more polymorphic relationships.

PR #205, opened 2020-05-11
First demo of resource abstraction

ryanwjackson:
The goal is to add an abstraction layer that allows us to have "opinionated" (aka not requiring ledger-specific knowledge) resources while also being able to use operations.

For example, one can represent an "expense" (depending on how you define it) in NetSuite in one of multiple record types (e.g. `Check`, `VendorPayment`, etc.).

A few notes:

1. This introduces a concept of `LedgerSync::Adaptors::Record`, which currently is just a child of `LedgerSync::Resource`.  The idea is that `Records` will be the ledger-specific resources.  Operations are designed to (though not required to) operate on records (aka ledger-specific objects).  **So now there is a distinction between `record` (ledger-specific resource) and `resource` (bundle resource).**
2. There are also "bundles" (better name welcome).  The bundles contain the "opinionated" part of LedgerSync.  In this PR, we start a `ModernTreasury` (MT) bundle, which will contain resources that MT uses.
3. Given that these resources may have attributes that are or are not included for a given ledger, they can be wrapped to create a "resource adaptor*." (see * notes below). You will need one resource adaptor per ledger you want to support with a given resource.
4. You can modify a resource adaptor class's `resource_module_str` method to say map an `Expense` resource to a `VendorPayment` resource.
5. There is a case to make that records do not need to define attributes and just let the operations handle validations.  After thinking about this (@jozefvaclavik this is especially for you given our talk), if we get rid of attributes, then records really do not represent anything.  The helper methods (like `name`) can all be done in serializers.  You will either be using a resource, resource adaptor, or just any object that defines the methods the operation contract expect.  But I find it hard to get rid of them, as they at least give you something tangible to operate on when interfacing with an operation using a record (not resource or resource adaptor).

* I use the term "adaptor" because that's technically what these are.  They modify the interface of an object to be usable in a different context.  I vote (though a discussion for a separate thread) that we rename our current "adaptors" folder/scope to "ledgers" to reduce confusion and be more specific.


lib/ledger_sync/adaptors/netsuite/records/customer.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module LedgerSync
+  module Adaptors
+    module NetSuite
+      module Records
+        class Customer < NetSuite::Record
+          attribute :email, type: LedgerSync::Type::String
+          attribute :companyName, type: LedgerSync::Type::String
```
ryanwjackson, line 9: I decided to follow the same convention of the ledger in terms of naming record attributes.  I figured this would be helpful if you were looking at the ledger documentation for attribute help.

SeanBolt:
@ryanwjackson I like this approach. I also like the idea of renaming our current `adaptors` folder/scope to `ledgers`.

PR #206, opened 2020-05-12
Rename adaptors to ledgers and connections

ryanwjackson:
Note, this is a breaking change, so we should definitely do a minor bump.


ryanwjackson:
@SeanBolt can you QA against Platform tests once you have made the naming changes?  Just want to test as much as possible.

README.md
```diff
@@ -32,16 +32,16 @@ $ gem install ledger_sync
...
 
 The code may look like the following:
 
 ```ruby
-# First we create an adaptor, which is our connection to a ledger.
+# First we create an connection, which is our connection to a ledger.
 # Each ledger may require different keys, so check the
 # documentation below for specifics.
-adaptor = LedgerSync::Adaptors::QuickBooksOnline::Adaptor.new(
+connection = LedgerSync::Ledgers::QuickBooksOnline::Ledger.new(
```
jozefvaclavik, line 44: I'm not sure I'm fan of this naming convention. Why not to name it `LedgerSync::Ledgers::QuickBooksOnline::Connection.new(`

README.md
```diff
@@ -145,30 +145,30 @@ customer_class_2.resource_attributes[:bar].type # => #<LedgerSync::Type::Boolean
...
 You can now use these custom resources in operations that require custom attributes.
 
-### Adaptors
+### Ledgers
 
-Adaptors are ledger-specific ruby objects that contain all the logic to authenticate to a ledger, perform ledger-specific operations, and validate resources based on the requirements of the ledger.  Adaptors contain a number of useful objects:
+Ledgers are ledger-specific ruby objects that contain all the logic to authenticate to a ledger, perform ledger-specific operations, and validate resources based on the requirements of the ledger.  Ledgers contain a number of useful objects:
 
-- adaptor
+- connectn
```
SeanBolt, line 152: This should be `client`

lib/ledger_sync/ledgers/quickbooks_online/client.rb
```diff
@@ -4,9 +4,9 @@
 require_relative 'dashboard_url_helper'
 
 module LedgerSync
-  module Adaptors
+  module Ledgers
     module QuickBooksOnline
-      class Adaptor < LedgerSync::Adaptors::Adaptor
+      class Client < Ledgers::Client
```
SeanBolt, line 9: Does this still work without the LedgerSync module reference?

spec/ledgers/netsuite/account/searcher_spec.rb
```diff
@@ -4,7 +4,7 @@
 
 support 'netsuite/shared_examples'
 
-RSpec.describe LedgerSync::Adaptors::NetSuite::Vendor::Searcher do
+RSpec.describe LedgerSync::Ledgers::NetSuite::Account::Searcher do
```
SeanBolt, line 7: Why do we move some of these specs from `Vendor`/`Currency` to `Account`?

lib/ledger_sync/ledgers/quickbooks_online/client.rb
ryanwjackson: Up!

PR #207, opened 2020-05-13
Cast NS account_types from camel case to snake case

SeanBolt:
Casting NS account_type values to snake case in order to be consistent across ledgers.


SeanBolt:
Closing in favor of #212

PR #209, opened 2020-05-18
Use new serializers everywhere

ryanwjackson:
Converted all the old `LedgerSerializers` to the new serializers.


lib/ledger_sync/deserializer.rb
```diff
@@ -68,6 +68,30 @@ def self.attribute_class
     def self.attributes
       @attributes ||= Serialization::DeserializerAttributeSet.new(serializer_class: self)
     end
+
+    def self.references_one(resource_attribute, args = {}, &block)
```
jozefvaclavik, line 72: Thanks. I was missing these last week 🍻

PR #211, opened 2020-05-21
Add NS customer deposit record

SeanBolt:


lib/ledger_sync/ledgers/netsuite/client.rb
```diff
@@ -63,7 +63,7 @@ def get(**keywords)
         def ledger_resource_path(args = {})
           resource = args.fetch(:resource, nil)
 
-          ret = resource.class.resource_type.to_s # This can be turned into a case statement if we need to override
+          ret = resource.class.resource_type.to_s.camelcase # This can be turned into a case statement if we need to override
```
SeanBolt, line 66: @ryanwjackson I did this as the api endpoint is `/customerDeposit`, whereas before it was hitting `/customer_deposit`, is this ok or is there a better way of doing this?

ryanwjackson: As long as this works for other NS routes.  We also have the `ledger_resource_type` overrides class methods on the Client.  It just isn't used in this function.  We could use it here (see the QBO Client for an example).

lib/ledger_sync/ledgers/netsuite/resources/customer_depost.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
```
ryanwjackson, line 1: This file name should be `customer_deposit.rb` (missing the `i`).

lib/ledger_sync/ledgers/netsuite/type/deserializer/account.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
```
ryanwjackson, line 1: If we follow the naming convention in [the lib root](https://github.com/LedgerSync/ledger_sync/tree/master/lib/ledger_sync/serialization/type), this file would be `lib/ledger_sync/ledgers/netsuite/type/deserializer_account_type.rb`

lib/ledger_sync/ledgers/netsuite/type/deserializer/customer.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
```
ryanwjackson, line 1: Naming convention see other type above.

lib/ledger_sync/ledgers/netsuite/resources/customer_depost.rb
SeanBolt: Great catch!

lib/ledger_sync/ledgers/netsuite/type/deserializer/account.rb
SeanBolt: I can change all the NS types to follow this pattern

SeanBolt: @ryanwjackson I'll leave this as it is for this PR as it follows the existing naming pattern and will follow up with a second PR to adjust all the NS types

PR #212, opened 2020-05-22
Add NS Account Type Mapping

SeanBolt:
We have determined that NS has a static list of account types, closing #207 in favor of this approach


lib/ledger_sync/ledgers/netsuite/resources/account.rb
```diff
@@ -6,9 +6,61 @@ module LedgerSync
...
+          'accounts_payable' => 'liability',
+          'credit_card' => 'liability',
+          'long_term_liability' => 'liability',
+          'other_current_liability' => 'liability',
+          'income' => 'revenue',
+          'other_income' => 'revenue',
+          'deferred_expense' => 'asset',
+          'unbilled_recievable' => 'assets',
+          'deferred_revenue' => 'liability'
+        }.freeze
```
SeanBolt, line 51: @ryanwjackson I copied this from the QBO account resource but it's not being used there either. How would we use this to populate the account classification values?

ryanwjackson: These are QBO specific, so they won't belong here.  

These are used in the resource attribute types (see `Type::StringFromSet`) and then again in the de/serializers for the respective resoruces.

ryanwjackson: @SeanBolt after some digging, I looked through the logs and found this: https://github.com/LedgerSync/ledger_sync/blob/master/lib/ledger_sync/ledgers/quickbooks_online/deserializer.rb#L16

It looks like a typo `TYPES` vs. `TYPE`.  I'll let you fix that here.  But I also believe this should be moved to the deserializer anyway.  I believe it can handle this logic in a block or a type.

Also, would be good to get a red/green test on this, because I'm guessing our request bodies always return a value and skip this block.

PR #220, opened 2020-05-29
Add NS Subsidiary Searcher

SeanBolt:


lib/ledger_sync/ledgers/netsuite/subsidiary/searcher.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module LedgerSync
+  module Ledgers
+    module NetSuite
+      class Subsidiary
+        class Searcher < Searcher
```
jozefvaclavik, line 7: I've seen this in other NS searcher classes, but this seems like something can go wrong. It should be
```suggestion
        class Searcher < NetSuite::Searcher
```

ryanwjackson: It is actually OK as it looks for a searcher in the most specific scope and then out.  So it doesn't see this one yet because it doesn't exist.  Though my preference is to have the `NetSuite::` just to make it more obvious.

PR #223, opened 2020-05-30
Remove entity_id attributes on NS customer/vendor

SeanBolt:
By setting the `entity_id` to the same value as `external_id` the records were displaying in NS with the name field containing the `entity_id`. By removing this attribute the name gets set to the `company_name`.

image


ryanwjackson:
@tatecarr does this ring any bells for you?

I'm curious if we should also be setting some other field so the name is displayed correctly.  And is there any downside to not setting `entity_id`?

PR #226, opened 2020-06-01
Use Reference::Serializer and add Memo to JELI

jozefvaclavik:
- Adds `memo` to `JournalEntryLineItem`
- Fixes [bad cleanup](https://github.com/LedgerSync/ledger_sync/pull/217/commits/79086009490538414582e31501701bedd3d73c4f#diff-485a770c5d3b228428e5976101b778ddL17)
- Removes `Type::SerializeReferenceType` and `ledger_reference`.

We had 3 ways to set relationship on a serializer.

#### 1. `attribute` + `type` combination

```ruby
attribute :subsidiary, references_one :subsidiary, type: Type::SerializerReferenceType.new 
```
I dont like several things about this.
a) it uses `attribute`
b) it differs from deserializer where we use `references_one`

#### 2. shorthand `ledger_reference`.

```ruby
ledger_reference :subsidiary
```

I get it, its short. But we are already having one too many ways to set `attribute` by using `attribute`, `reference_one`, `mapping`, etc.

#### 3. specifying specific serializer in `references_one`
```ruby
references_one :subsidiary, serializer: Reference::Serializer
```
To me this is the cleanest way. It is clear it is reference and it is clear what serializer it's using.


ryanwjackson:
@jozefvaclavik I agree.  But I still think the shorthand is fine.  We could just make it set the serializer.  But I am definitely for being explicit and using the serializers over the type in this case.

PR #262, opened 2020-10-07
Fix gem vulnerability, update Client, and add test ledger

ryanwjackson:
We needed to update the kramdown version, which had some trickle-down dependencies (aka more justification for separating gems). It cascaded into needing to update rubocop, the `Client` core class (not a module), and adding the `TestLedger` for more accurately testing core.


spec/support/test_ledger/operation/full_update.rb
```diff
@@ -0,0 +1,42 @@
...
+          private
+
+          def find_resource
+            client.find(
+              path: ledger_resource_path
+            )
+          end
+
+          def operate
+            find_result_hash = find_resource.body[quickbooks_online_resource_type.to_s.classify]
```
jozefvaclavik, line 21: `quickbooks_online_resource_type`
```suggestion
            find_result_hash = find_resource.body[test_ledger_resource_type.to_s.classify]
```

spec/support/test_ledger/operation/create.rb
```diff
@@ -0,0 +1,27 @@
...
+    module TestLedger
+      class Operation
+        class Create
+          include TestLedger::Operation::Mixin
+
+          private
+
+          def operate
+            success(
+              resource: TestLedger::Customer.new(
```
jozefvaclavik, line 16: Wouldnt it be better to just return resource we pass to operation?

spec/support/test_ledger/operation/full_update.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+require_relative '../operation'
+
+module LedgerSync
+  module Ledgers
+    module TestLedger
+      class Operation
+        class FullUpdate
```
jozefvaclavik, line 9: Also I would say `FullUpdate` is very QBO specific. I would drop this completely and just leave `Update`

spec/support/test_ledger/customer/operations/update.rb
```diff
@@ -0,0 +1,26 @@
...
+            class Contract < LedgerSync::Ledgers::Contract
+              params do
+                required(:external_id).maybe(:string)
+                required(:ledger_id).filled(:string)
+                required(:PrimaryEmailAddr).maybe(:hash, Types::Reference)
+                optional(:DisplayName).filled(:string)
+                optional(:firstName).maybe(:string)
+                optional(:lastName).maybe(:string)
+                required(:PrimaryPhone).maybe(:hash, Types::Reference)
+                required(:subsidiary).maybe(:hash, Types::Reference)
```
jozefvaclavik, line 18: I'm pretty sure we dropped `subsidiary` from QBO as it doesnt exist there. Also why the difference in the camel case here?

ryanwjackson: Updated, but these weren't even being used.

amouxkin (approved):

PR #265, opened 2020-10-17
Require only LedgerSync namespace for ledgers

jozefvaclavik:
Right now `LedgerSync::Ledgers` is required for any ledger. We infer ledgername from module and expect `LedgerSync::Ledgers` to be there. This PR drops `LedgerSync::Ledgers` and requires only `LedgerSync`.

We should probably drop `LedgerSync` as well. Just wanted to do minimum necessary to make `LedgerSync::Xero` working.


ryanwjackson:
I'm not sure I follow. Should this also look at the `LedgerConfiguration` module or root path?

jozefvaclavik:
Currently the main gem expects every ledger to live under `LedgerSync::Ledgers` namespace. This PR make it to be only `LedgerSync` namespace.

Our endgame should be to allow ledger to use any namespace and still be able to use `LedgerSync.register_ledger(:name, { root_path: 'my/path/to/ledger' })` method.

jozefvaclavik:
I made some more changes and now it uses `base_module` to infer client and resource classes.

if we pass `base_module` to `LedgerSync.register_ledger`, it will use that instead of `LedgerSync` module namespace.

ryanwjackson (approved):

PR #267, opened 2020-11-01
Namespaced resources

jozefvaclavik:
As I was going through 1.5.2 I noticed that it doesn't work well with two cases

### Single level module name
If module was called `Xero` instead of `LedgerSync::Xero` (ie someone building their integration with LS), our `inferred_config` method didn't get the module right.

### Nested resources
I was thinking a lot how to do nested resources. For example `/api/customers` and also `/api/accounting/invoices`. Turned out it wasnt that much additional work. We load resources first through `resources/**/*.rb` so anything under `resources` as well as `resources/accounting` will get loaded first. Then we load everything else recursively.

So now you can have resource like `MyLedger::Customer` and `MyLedger::Accounting::Invoice`

The biggest issue was again inferring resource class. I made two changes
1. if we're inside of `Operation` it is easier to get resource by splitting on `::Operations::` instead of subtracting `base_module`.
2. to get serializer/deserializer we were using `resource_module_str` while we already had the inferred resource class.

I think its good convention to expect `MyModule::MyResource::Operations::*` as well as `MyModule::MyResource::Serializer` and `MyModule::MyResource::Deserializer`. Therefore we always have resource class above operations module and (de)serializer after resource class.

### DIFF
--- a/lib/ledger_sync/ledgers/mixins/infer_config_mixin.rb
+++ b/lib/ledger_sync/ledgers/mixins/infer_config_mixin.rb
@@ -6,9 +6,9 @@ module LedgerSync
       module InferConfigMixin
         module ClassMethods
           def inferred_config
-            @inferred_config ||= begin
-              return if name.nil?
+            return if name.nil?
 
+            @inferred_config ||= begin
               name_parts = name.split('::')
               name_parts_length = name_parts.count
 
--- a/lib/ledger_sync/ledgers/mixins/infer_config_mixin.rb
+++ b/lib/ledger_sync/ledgers/mixins/infer_config_mixin.rb
@@ -17,7 +17,7 @@ module LedgerSync
               name_parts_length.times do |i|
                 config = LedgerSync.ledgers.config_from_base_module(
                   base_module: Object.const_get(
-                    name_parts[0..(name_parts_length - i)].join('::')
+                    name_parts[0..(name_parts_length - 1 - i)].join('::')
                   )
                 )
                 break if config.present?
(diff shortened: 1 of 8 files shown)


lib/ledger_sync/ledgers/mixins/infer_config_mixin.rb
```diff
@@ -17,7 +17,7 @@ def inferred_config
               name_parts_length.times do |i|
                 config = LedgerSync.ledgers.config_from_base_module(
                   base_module: Object.const_get(
-                    name_parts[0..(name_parts_length - i)].join('::')
+                    name_parts[0..(name_parts_length - 1 - i)].join('::')
```
jozefvaclavik, line 20:
```
irb(main):009:0> name_parts.count.times {|i| puts name_parts[0..(name_parts.length - i)].join('::') }
MyLedger::Customer
MyLedger::Customer
=> 2
irb(main):010:0> name_parts.count.times {|i| puts name_parts[0..(name_parts.length - 1 - i)].join('::') }
MyLedger::Customer
MyLedger
=> 2
```

ryanwjackson:
@jozefvaclavik few comments:

1. Can you write specs so we guarantee all the edge cases that led to this?
2. Can you document in comments above the functions the intent, inputs, and outputs?
3. We should keep in mind these are inferred for ease-of-use, but ultimately, you should be able to specify this one-off somewhere (e.g. passed in when instantiating the object or overriding the methods). Is that possible here?

ryanwjackson:
Also @jozefvaclavik did you verify this works with the new Xero PR?

jozefvaclavik:
@ryanwjackson yeah, I tested it on Xero. All Xero tests are green.

ryanwjackson (approved):

ryanwjackson (approved):

PR #275, opened 2021-01-09
Resolve assign_attributes warning

jozefvaclavik:
When you make an instance of a resource in ruby 2.7 you get quite cryptic warning. This causes crash in ruby 3.0. The fix is quite simple. We have `assign_attributes` defined in `ResourceAttribute::Mixin` as well as in `Resource` class that uses this mixin. We should not use `**` around the attributes. I believe the double definition is needed to return `self` in the resource class (in mixin self is not same as in class).

I also added ruby 3.0.0 to travis 🤞🏻 But for ruby 3 we need to update jekyll gems.

### Ruby 2.7

```ruby
irb(main):001:1* class Whatever < LedgerSync::Resource
irb(main):002:1*   attribute :name, type: LedgerSync::Type::String
irb(main):003:0> end
=> #<LedgerSync::ResourceAttribute:0x00007fb3bd1327e8 @name=:name, @resource_class=Whatever, @type=#<LedgerSync::Type::String:0x00007fb3bd132748 @precision=nil, @scale=nil, @limit=nil>, @value=nil>
irb(main):004:0> Whatever.new(name: 'Yay')
/Users/jozefvaclavik/Developer/mt/ledger_sync/lib/ledger_sync/resource_attribute/mixin.rb:60: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/Users/jozefvaclavik/Developer/mt/ledger_sync/lib/ledger_sync/resource.rb:41: warning: The called method `assign_attributes' is defined here
=> #<Whatever:0x00007fb3bd1b9bf8 @resource_attributes=#<LedgerSync::ResourceAttributeSet:0x00007fb3bd1b9770 @attributes={:external_id=>#<LedgerSync::ResourceAttribute:0x00007fb3bd1b9630 @name=:external_id, @resource_class=LedgerSync::Resource, @type=#<LedgerSync::Type::ID:0x00007fb3bd1b94c8 @precision=nil, @scale=nil, @limit=nil>, @value=nil>, :ledger_id=>#<LedgerSync::ResourceAttribute:0x00007fb3bd1b9388 @name=:ledger_id, @resource_class=LedgerSync::Resource, @type=#<LedgerSync::Type::ID:0x00007fb3bd1b9360 @precision=nil, @scale=nil, @limit=nil>, @value=nil>, :name=>#<LedgerSync::ResourceAttribute:0x00007fb3bd1b9310 @name=:name, @resource_class=Whatever, @type=#<LedgerSync::Type::String:0x00007fb3bd1b9298 @precision=nil, @scale=nil, @limit=nil>, @value="Yay">}, @references=[], @references_one=[], @references_many=[], @resource=Whatever>, @external_id=nil, @ledger_id=nil, @name=nil, @mutations_from_database=#<ActiveModel::ForcedMutationTracker:0x00007fb3bd1b89b0 @attributes=#<Whatever:0x00007fb3bd1b9bf8 ...>, @forced_changes={"name"=>nil}, @finalized_changes=nil>>
irb(main):005:0>
```

### Ruby 3.0
```ruby
irb(main):001:1* class Whatever < LedgerSync::Resource
irb(main):002:1*   attribute :name, type: LedgerSync::Type::String
irb(main):003:0> end
=> #<LedgerSync::ResourceAttribute:0x00007f83f4a78210 @name=:name, @resource_class=Whatever, @type=#<LedgerSync::Type::String:0x00007f83f4a78120 @precision=nil, @scale=nil, @limit=nil>, @value=nil>
irb(main):004:0> Whatever.new(name: 'Yay')
Traceback (most recent call last):
        5: from ./bin/console:15:in `<main>'
        4: from (irb):4:in `<main>'
        3: from (irb):4:in `new'
        2: from /Users/jozefvaclavik/Developer/mt/ledger_sync/lib/ledger_sync/resource_attribute/mixin.rb:60:in `initialize'
        1: from /Users/jozefvaclavik/Developer/mt/ledger_sync/lib/ledger_sync/resource.rb:41:in `assign_attributes'
ArgumentError (wrong number of arguments (given 1, expected 0))
irb(main):005:0>
```
---

### Update:
This went down the rabbit hole. Apparently you need to be [more explicit](https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/) about use of keyword arguments in ruby 3. This makes tests green in ruby 2.7 as well as 3.0.


ryanwjackson:
Is this just future ruby versions? Does it make sense to wait until you get the docs moved out to do this?

ryanwjackson (approved):
Made a few minor edits, but lgtm.

PR #299, opened 2022-05-31
Bump minimum ruby version support

ryanwjackson:
In order to fix dependebot and other security flaws, we need to deprecate support for older ruby versions.


amouxkin (approved):

amouxkin (approved):

amouxkin (approved):

ledger_sync.gemspec
```diff
@@ -50,6 +45,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency('simplecov-lcov')
   spec.add_development_dependency('webmock', '>= 0')
   spec.add_runtime_dependency('activemodel', '>= 0')
+  spec.add_runtime_dependency('dotenv')
```
jozefvaclavik, line 48: Any reason for this being required at runtime?

ryanwjackson: Yes, because `.env` files can be used in production environments too, and this ensures they are overwritten if need be. Though I'm not sure if it needs to be in core vs. QBO where OAuth 2.0 requires updating that file in at least dev.

jozefvaclavik: For example we dont use `dotenv` in production, but this forces us to install it.

ryanwjackson: This is no longer in this PR as far as I can tell. Want to approve?

jozefvaclavik (approved):

PR #302, opened 2022-07-22
Update openssl dependency version

metalogical:
This gem seems to be fully compatible with openssl 3.0.0, and openssl 2.2.0 has issues building on Ventura, so allow versions `< 4`


ryanwjackson:
@metalogical you will need to remove the ruby version `2.5.8` from https://github.com/LedgerSync/ledger_sync/blob/3fb4826a134a42318514e188bdee6dcb1fd003db/.github/workflows/gem-workflow.yml or use a lower version of openssl in the committed gemlock.

ryanwjackson:
Actually, I have a separate PR that may solve this: https://github.com/LedgerSync/ledger_sync/pull/303

I'm discussing with the other maintainers, but once approved, we can merge and then rebase this PR on top.

PR #307, opened 2022-09-19
feat: allow request to use custom faraday client instance

KostasKostogloy:
Using global Faraday instance/configuration is a bit restrictive. 
You can't:
- have different configuration for different APIs (adapter, request timeout etc...) 
- pass your own faraday middleware (we use a middleware to log requests & responses)

For example you might want to use Faraday's `:test` [adapter](https://lostisland.github.io/faraday/adapters/testing) to mock requests during testing.

With this PR you can pass your custom configured Faraday client instance on your ledger's client and from there pass it to your request.

If you don't pass your own client, it defaults to creating an instance (which uses your global faraday default options).

### DIFF
--- a/lib/ledger_sync/ledgers/request.rb
+++ b/lib/ledger_sync/ledgers/request.rb
@@ -16,6 +16,7 @@ module LedgerSync
         @method = args.fetch(:method, nil)
         @params = args.fetch(:params, {})
         @url = args.fetch(:url, nil)
+        @faraday_client = args.fetch(:faraday_client, Faraday.new)
       end
 
       def perform
@@ -23,7 +24,7 @@ module LedgerSync
 
         url_with_params = Util::URLHelpers.merge_params_in_url(params: params, url: url)
 
-        faraday_response = Faraday.send(method, url_with_params) do |req|
+        faraday_response = @faraday_client.send(method, url_with_params) do |req|
           req.headers = headers
           req.body = body.to_json unless body.nil?
         end


jozefvaclavik (approved):

lib/ledger_sync/ledgers/request.rb
```diff
@@ -16,14 +16,15 @@ def initialize(args = {})
         @method = args.fetch(:method, nil)
         @params = args.fetch(:params, {})
         @url = args.fetch(:url, nil)
+        @faraday_client = args.fetch(:faraday_client, Faraday.new)
```
ryanwjackson, line 19: Should this be `Faraday.new` or `Faraday`? The old code would suggest `Faraday` but maybe these are the same. Can you verify?

KostasKostogloy: They're not the same. `Faraday.new` creates a `connection` (don't like the term but this is what [Faraday docs](https://lostisland.github.io/faraday/usage/) use) that uses Faraday's default settings as they were at that point in time. 

```ruby
Faraday.default_adapter = Faraday::Adapter::Typhoeus
Faraday.adapter
=> Faraday::Adapter::Typhoeus

connection = Faraday.new
connection.adapter
=> Faraday::Adapter::Typhoeus

Faraday.default_adapter = :test
Faraday.adapter
=> Faraday::Adapter::Test

connection.adapter
=> Faraday::Adapter::Typhoeus
```

If you want to keep backwards compatibility to 100%, you'd use `Faraday` instead of `Faraday.new`.
But the only difference it makes is if you dynamically change Faraday default configuration while running the app (instead of just once when you initialise it), which is very problematic on its own.

We could also opt to use `connection` rather than `faraday_client` as the variable name, if we want to stay more inline with Faraday's own terminology.

ryanwjackson: To confirm `Faraday.new` accepts `post` or other method names just like `Faraday`?

KostasKostogloy: @ryanwjackson that's correct

ryanwjackson (approved):

PR #318, opened 2023-06-01
Fix: Namespaced Resource key

jamshi:
**Why ?**

When we have a genuine scenario to have same class name under different folder, current `resource_module_str` method prevents to have same class name.
Screenshot 2023-06-07 at 11 39 16 AM

Look at the above structure. Here we have resource `shipping_address` under `marketplace` and root `resources` folder. With the current `resource_module_str` implementation of just taking the class name we end up in below error.

```bash
`rescue in block (2 levels) in require': There was an error while trying to load the gem 'desertcart'. (Bundler::GemRequireError)
Gem Load Error is: Resource key shipping_address already exists.
Backtrace for gem load error is:                    
/Users/jamsheed/.asdf/installs/ruby/3.2.0/lib/ruby/gems/3.2.0/bundler/gems/ledger_sync-095cf8d36086/lib/ledger_sync/util/mixins/resource_registerable_mixin.rb:10:in `register_resource'
/Users/jamsheed/.asdf/installs/ruby/3.2.0/lib/ruby/gems/3.2.0/bundler/gems/ledger_sync-095cf8d36086/lib/ledger_sync/resource.rb:73:in `inherited'        
/Users/jamsheed/sandbite/desertcart-ruby/lib/desertcart/resources/shipping_address.rb:4:in `<module:Desertcart>'
/Users/jamsheed/sandbite/desertcart-ruby/lib/desertcart/resources/shipping_address.rb:3:in `<main>'
/Users/jamsheed/.asdf/installs/ruby/3.2.0/lib/ruby/gems/3.2.0/gems/bootsnap-1.14.0/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:32:in `require'
``` 
**Fix**
Rather than taking the class name as key, this PR attempts to have a namespaced key by extracting everything after the `base_module` name to key.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,76 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "master", develop ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ "master" ]
+  schedule:
+    - cron: '38 21 * * 6'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ${{ (matrix.language == 'swift' && 'macos-latest') || 'ubuntu-latest' }}
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'ruby' ]
+        # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python', 'ruby' ]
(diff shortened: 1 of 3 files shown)


ledger_sync.gemspec
```diff
@@ -53,7 +53,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency('faraday_middleware', '>= 0')
   spec.add_runtime_dependency('fingerprintable', '>= 1.2.1')
   spec.add_runtime_dependency('nokogiri', '>= 0')
-  spec.add_runtime_dependency('openssl', '~> 2.2.0')
+  spec.add_runtime_dependency('openssl', '~> 3.0')
```
ryanwjackson, line 56: I think we should just do `> 2.2.0` for now. This could be a breaking change for some. There's some wonkiness with the openssl version, which is why it's even included here. I haven't found the root cause, but it seems related to the ruby 2 > 3 migration. So I don't want to disallow openssl 2.2 but we can open it up to allow 3+. Does that work?

ryanwjackson (changes requested):
@jozefvaclavik can you actually separate out these into two PRs, as they aren't related. I'm OK to approve both, but I want to be able to rollback in isolation if needed. I'm concerned the module changes may have downstream impacts, though I think this looks OK.

ryanwjackson (approved):

PR #322, opened 2024-08-29
Ensure references_many no nil value

jkeuleya:


lib/ledger_sync/resource_attribute.rb
```diff
@@ -72,5 +74,11 @@ def will_change?(val)
       assert_valid(value: val)
       value != type.cast(value: val)
     end
+
+    def ensure_references_many_value_presence(args)
+      if references_many? && args.fetch(:value) == nil
+        args[:value] = []
+      end
+    end
```
jkeuleya, line 82: Basically, when an array is not received in the answer (in my exemple it's amazon business api), `references_many` is `nil`. Its value is `nil` too. I ensure here that if we meet this specific case, we set an empty array instead `[]`

lib/ledger_sync/resource_attribute.rb
```diff
@@ -72,5 +74,11 @@ def will_change?(val)
       assert_valid(value: val)
       value != type.cast(value: val)
     end
+
+    def ensure_references_many_value_presence(args)
+      if references_many? && args.fetch(:value).nil?
+        args[:value] = []
+      end
+    end
```
jkeuleya, line 82: Basically, when an array is not received in the answer (in my case it's amazon business api), `references_many` is `nil`. Its value is `nil` too. I ensure here that if we meet this specific case, we set an empty array instead `[]`.

Why here? Because after, comes the `.set_attribute(key, value)` and it can't be `nil`.

I checked files and methods before arriving here. I followed manually the path taken by the code execution. I think that this file is the best place to check value's presence. Other files are even more generic and not appropriate imo.

jozefvaclavik:
Hey @jkeuleya. Thx for working on this. I'm afraid that this PR is trying to fix it on a wrong layer. Lets try something else here. From what I tracked down is that during deserialization we get the value from has and then we `cast` it on type in https://github.com/LedgerSync/ledger_sync/blob/master/lib/ledger_sync/serialization/deserializer_attribute.rb#L36 from there we exit early in `cast` method because value is `nil` in https://github.com/LedgerSync/ledger_sync/blob/master/lib/ledger_sync/type/value_mixin.rb#L25. So the `DeserializerReferencesManyType` and its `cast_value` method are never called. As this is specific just for this type, that feels like its the right place to handle it.

Lets remove the exit for `nil` value in `cast` method as I believe the type is responsible how the value (even `nil`) needs to be casted. And then in `DeserializerReferencesManyType` instead of returning `nil` if value is `nil`, lets return empty array `[]` instead.

Let me know if this approach can work and resolves the issue.

ryanwjackson:
@jkeuleya can you also write a test for this so we ensure we are capturing the expected behavior? That would help me know what you're ultimately trying to achieve.

If I am understanding correctly, I'm concerned that this change would be converting `nil` to `[]` which I don't think is the correct behavior.

It may be easiest to just have a mock of the Amazon Business API response that is failing to create a feature test.

jkeuleya:
Thank you everyone for your time. I updated the description with an example.
I'll provide you today or Monday a spec if possible and go through @jozefvaclavik 's proposition !

spec/deserializer/acceptance_artifact_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+
+class Package
+  class Deserializer < LedgerSync::Deserializer
+    attribute :blabla, hash_attribute: :blabla
+  end
+end
+
+class AcceptanceArtifact < LedgerSync::Resource
+  attribute :identifier, type: LedgerSync::Type::String
+  references_many :packages, to: Package
```
ryanwjackson, line 18: What if we add a param called `default` to `references_many` so that you could do `default: []`? I worry that your implementation here is going to be a breaking change. Unless I am missing something, I believe one could argue either way that an API not returning a parameter should result in an empty list or `nil`. So I think we should make it configurable to make this change backwards compatible.

@jozefvaclavik do you agree?

jkeuleya: It's exactly what I searched at the beginning, a `default`. It was the easy solution but it doesn't exist yet. So yes, it could've been a solution.

For my first implementation, yes, it was a naive approach, I didn't take time to check the whole code base. My goal was to even avoid `default` param and just do a presence check.

But still, need to think about the implementation of a `default` while not creating a breaking change.. :thinking:

lib/ledger_sync/serialization/type/deserializer_references_many_type.rb
```diff
@@ -11,7 +11,7 @@ def cast_value(args = {})
           value                  = args.fetch(:value)
           resource               = args.fetch(:resource)
 
-          return if value.nil?
+          return [] if value.nil?
```
jkeuleya, line 14: In this proposition (thanks @jozefvaclavik for your time and your research), we're being more precise/meticulous.

We're working directly on the references_many file.

Nonetheless, we also touched the `value_mixin` file (see below) which could lead to edge cases. I cannot guarantee here. I did several tests with several values and types. I put some breaking points in the code execution trying to understand and check. But still.

lib/ledger_sync/type/value_mixin.rb
```diff
@@ -22,7 +22,6 @@ def assert_valid(args = {})
       # Do not override this method.  Override private method cast_value
       def cast(args = {})
         assert_valid(args)
-        return nil if args.fetch(:value).nil?
```
jkeuleya, line 25: We remove here this responsibility to `cast` method.

PR #323, opened 2024-09-04
References many default

ryanwjackson:


lib/ledger_sync.rb
```diff
@@ -15,6 +15,7 @@
 require 'faraday'
 require 'tempfile'
 require 'pd_ruby'
+require 'pry'
```
ryanwjackson, line 18: Need to remove

lib/ledger_sync/resource_attribute.rb
```diff
@@ -37,16 +37,16 @@ def assert_valid(args = {})
       value = args.fetch(:value)
 
       type.assert_valid(value: value)
-    rescue Error::TypeError::ValueClassError
-      raise ResourceAttributeError::TypeError.new(
-        attribute: self,
-        resource_class: resource_class,
-        value: value
-      )
+    # rescue Error::TypeError::ValueClassError
```
ryanwjackson, line 40: Uncomment

ryanwjackson:
@jozefvaclavik this is more in line with what I'm thinking for defaults, but I have one failing coverage test.

